PR #38, opened 2026-05-08
Add runner heartbeat protocol and orchestration timeout

noe-charmet:
Make runner liveness reliable so a disappeared runner can no longer park `jobOrchestration` on its signal forever and leave `running_jobs` rows behind. Closes the highest-priority production gap from `TODOS.md`.

Four interlocking pieces:

- **`POST /runners/jobs/:jobId/heartbeat`** — refreshes `last_heartbeat_at` and returns whether the orchestration has requested cancellation. Mirrors the existing `complete-job` route, reuses runner-token auth.
- **Runner heartbeat scheduler** (`apps/runner/src/heartbeat-loop.ts`) — single-flight, `setTimeout`-chained, with a 30s max-stale guard that aborts the in-flight HTTP call before scheduling the next tick. At most one call in flight under any API latency. `cancel:true` and HTTP 404 abort the per-job `AbortController`, which kills the running shell **and its grandchildren** via `spawn(..., {detached:true})` + `process.kill(-pid, 'SIGKILL')`. Module-level `activeProcess` removed; SIGINT now routes through the same controller.
- **Stuck-job-detector cron** — Temporal cron (`* * * * *`, 180s threshold) running `detectAndFailStuckJobs`. Uses a single atomic `DELETE…RETURNING` with the cutoff predicate inside the `WHERE` clause, so a heartbeat refresh between the iteration `SELECT` and the per-row `DELETE` leaves the live row intact.
- **`jobOrchestration` timeout** — `condition()` now takes `JOB_MAX_DURATION = '60 minutes'`. On timeout, `requestJobCancellationActivity` (bounded retry: `maximumAttempts:1`, `scheduleToCloseTimeout:'10s'`) sets `cancellation_requested_at` so the runner picks it up on its next heartbeat; failure is logged via workflow-side `log.warn` and the job still transitions to failed with `output.reason = 'job_timeout'`.

## Why this shape

`condition(timeout)` and `cronSchedule` are Temporal-native — no custom polling or continueAsNew loop. Atomic `DELETE…RETURNING` with the cutoff predicate is what makes the "exactly one DELETE wins" invariant hold under heartbeat refresh + concurrent runner `/complete`. Module decoupling preserved: workflows-side activities are thin wrappers, all SQL stays inside `@shipfox/api-runners`. Public surface added: `requestJobCancellation` and `detectAndFailStuckJobs` only.

The plan was reviewed via `/plan-eng-review` and `/codex review`. Codex caught six real concurrency / retry-policy bugs the in-house review missed (atomic DELETE race, default activity retries, max-stale needing HTTP-level abort, `/complete` 404 handling, reason field never reaching the workflow, missing process-group kill); all six are fixed in the diff. Plan + review notes live at `.claude/plans/system-instruction-you-are-working-fluttering-bubble.md`.

## Backwards-incompatible bits worth flagging in review

- **Schema change folded into the existing `0000_initial.sql` migration**, not a new 0002 file (per "drop the local DB and re-migrate" workflow — nothing has shipped yet). Adds `cancellation_requested_at` (nullable) and an index on `last_heartbeat_at`. **Local DBs need to be dropped + remigrated.**
- **`completeJob` refactored** to share a private `finalizeRunningJob` helper. Same external behavior; also fixes a pre-existing SELECT-then-DELETE race in the existing complete path (loser of a concurrent finalize could write a stale outbox event).
- **`WorkflowStart` interface** in `@shipfox/node-module` gains an optional `cronSchedule?: string` field forwarded to `temporalClient.workflow.start`. No-op for existing callers.
- **`runner-token` auth path: `runners-running-jobs` now exposes `cancellation_requested_at`** on the response of `POST /runners/jobs/:id/heartbeat`. Fresh DTO; no change to existing endpoints.

## Tests

All 72 test packages pass. New coverage:

- `libs/api/runners/src/db/jobs.test.ts` — 15 new tests including the codex F1 race scenario (heartbeat refresh between iteration SELECT and atomic DELETE).
- `libs/api/runners/src/presentation/routes/heartbeat.test.ts` — 5 tests (200/cancel:false, 200/cancel:true, 404 unknown job, 404 wrong token, 401 unauth).
- `libs/api/workflows/src/temporal/workflows/job-orchestration-timeout.test.ts` — uses `TestWorkflowEnvironment.createTimeSkipping()` to verify the 60-minute timeout fires and that the failure path proceeds even when `requestJobCancellationActivity` throws.
- `apps/runner/src/heartbeat-loop.test.ts` — 7 unit tests with fake timers covering single-flight, max-stale abort, cancel:true/404 propagation, transient error retry, `stop()`.
- `apps/runner/src/run-step.test.ts` — process-group kill test that spawns a grandchild, aborts, and verifies `ESRCH` on `process.kill(grandchildPid, 0)`.

## Closes / refs

Closes the "Add job execution timeout + stuck runner detection" entry in `TODOS.md`. Adds five new TODOs:
- Per-job-type configurable timeout
- Bulk DELETE…RETURNING + bulk outbox in `detectAndFailStuckJobs`
- `cancelRun(runId)` + DAG-wide cancellation propagation
- Calibrate stuck-job threshold + add `jobs_failed_by_detector` metric
- Move stuck-job-detector cron + activity into a runners-module Temporal worker

---
## Summary by cubic
Adds a heartbeat protocol, a stuck-job detector, and a 60-minute job timeout so vanished runners no longer park workflows or leave `running_jobs` behind. Jobs now always reach a terminal state and runners get cancel signals via an event-driven path.

- **New Features**
  - `POST /runners/jobs/:jobId/heartbeat` refreshes `last_heartbeat_at` and returns `{cancel}`; uses runner-token auth.
  - Runner heartbeat loop (`apps/runner`): single-flight scheduler with a max-stale guard (10s default) that aborts in-flight HTTP; `cancel:true` or HTTP 404 aborts the job and kills the process group.
  - Stuck-job detector: runners-owned Temporal cron (`* * * * *`, `runners-maintenance` queue, 180s threshold) finalizes stale rows with an atomic `DELETE … RETURNING`, emitting `runners.job.completed` with reason `runner_disappeared`.
  - Job orchestration timeout: waits up to 60 minutes; on timeout `failJobAsTimedOutActivity` atomically marks the job failed, sets `jobs.timed_out_at`, and emits `workflows.job.timed_out`. A new runners subscriber sets `cancellation_requested_at` so the runner cancels; the workflow returns `failed` with `output.reason = 'job_timeout'`.
  - Safer finalization: race-free `finalizeRunningJob` replaces SELECT-then-DELETE; runner completion 404s are treated as already finalized, and late completion signals are dropped if the workflow already ended.
  - Internal split: orchestration lives in `@shipfox/api-runners` core (`completeJob`, `detectAndFailStuckJobs`); DB layer exposes primitives (`finalizeRunningJob`, `findStuckJobs`, `recordHeartbeat`, `requestJobCancellation`). The runners module registers its own Temporal worker and cron.
  - Package API: `@shipfox/api-runners` now exports only `runnersModule` and `enqueueJob`; other internals are no longer public.

- **Migration**
  - Drop and re-run local migrations: `runners_running_jobs` adds `cancellation_requested_at` and an index on `last_heartbeat_at`; `workflows.jobs` adds `timed_out_at` in `0000_initial.sql`.

<sup>Written for commit a8fbba3affc5028634412f11b8f6fe0508805b82. Summary will update on new commits.</sup>


apps/runner/src/api-client.ts
```diff
@@ -38,4 +42,16 @@ export async function completeJob(params: {
   });
 }
 
+export async function heartbeat(
+  jobId: string,
+  options: {signal?: AbortSignal} = {},
+): Promise<{cancel: boolean}> {
```
noe-charmet: The API return type should be read from the related DTO package. Same goes for request body/params type. Applies to all requests in this file.

apps/runner/src/config.ts
```diff
@@ -5,4 +5,9 @@ export const config = createConfig({
   SHIPFOX_POLL_INTERVAL_MS: num({default: 5000}),
   SHIPFOX_POLL_MAX_INTERVAL_MS: num({default: 30000}),
   SHIPFOX_RUNNER_TOKEN: str({default: 'static-poc-token'}),
+  // Heartbeat tick interval. Must be << stuck-job threshold (180s server-side).
+  SHIPFOX_HEARTBEAT_INTERVAL_MS: num({default: 10_000}),
+  // Max time a single in-flight heartbeat may stay outstanding before the loop
+  // aborts it and schedules the next tick. Bounds overlap under a hung API.
+  SHIPFOX_HEARTBEAT_MAX_STALE_MS: num({default: 30_000}),
```
noe-charmet: We can reduce this to 10s

apps/runner/src/heartbeat-loop.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  return new HTTPError(response, request, options);
+}
+
+describe('startHeartbeatLoop', () => {
+  test('schedules first tick at intervalMs and calls heartbeat with the job id', async () => {
+    heartbeatMock.mockResolvedValue({cancel: false});
+    const ac = new AbortController();
+
+    const handle = startHeartbeatLoop('job-1', ac, {intervalMs: 100, maxStaleMs: 1000});
+
```
noe-charmet, line 31: Do we need to wait for at least one event loop tick here to confirm it does not immediately fire ?

apps/runner/src/run-step.test.ts
```diff
@@ -68,6 +71,47 @@ describe('executeRunStep', () => {
...
+
+    // Give the shell a moment to actually spawn before we abort.
+    await new Promise((r) => setTimeout(r, 50));
+    ac.abort();
+
+    const result = await promise;
+    expect(result.success).toBe(false);
+  });
+
+  it('kills the entire process group on abort, including grandchildren (codex F9)', async () => {
```
noe-charmet: Don't mention codex f9 it's not relevant for future readers

apps/runner/src/run-step.ts
```diff
@@ -29,46 +32,40 @@ export function executeRunStep(step: JobPayloadStepDto): Promise<StepResult> {
...
   return new Promise((resolve) => {
     const shell = findShell();
     const args =
       shell === 'bash'
         ? ['--noprofile', '--norc', '-eo', 'pipefail', scriptPath]
         : ['-e', scriptPath];
 
+    // detached: true puts the child in its own process group so we can SIGKILL
+    // the whole group on abort — a shell script that spawns Node, etc, would
+    // otherwise leave grandchildren orphaned (codex F9). We do NOT call
```
noe-charmet: Same on codex f9

apps/runner/src/runner.ts
```diff
@@ -57,11 +50,52 @@ export async function startRunner(): Promise<void> {
...
+    logger().info({jobId: job.job_id, status: result.status}, 'Job execution finished');
+  } catch (execError) {
+    logger().error({err: execError, jobId: job.job_id}, 'Job execution failed');
+    result = {status: 'failed', output: String(execError)};
+  } finally {
+    heartbeatLoop.stop();
+    if (currentJobAbortController === ac) currentJobAbortController = undefined;
+  }
+
+  // Codex F5: separate try/catch around the completion report. A 404 here is
```
noe-charmet: Don't mention CODEX F% it's unhelpful to the reader

libs/api/runners/src/db/jobs.test.ts
```diff
@@ -187,3 +194,234 @@ describe('completeJob', () => {
...
   });
 });
+
+describe('recordHeartbeat', () => {
+  let workspaceId: string;
+  let runnerTokenId: string;
+
+  beforeEach(async () => {
+    await db().execute(
+      sql`TRUNCATE runners_pending_jobs, runners_running_jobs, runners_runner_tokens, runners_outbox CASCADE`,
```
noe-charmet: This should be in the modules globalSetup. Check how we do for example here libs/api/projects/test/globalSetup.ts

This is vaulable for all tests in this module

libs/api/runners/src/db/jobs.ts
```diff
@@ -70,44 +70,157 @@ export async function claimJob(params: {
...
+  /** 'throw' (route paths) or 'noop' (detector paths). */
+  onMissing: 'throw' | 'noop';
+}
+
+/**
+ * Atomic finalize: a single DELETE...RETURNING with all guard predicates folded
+ * into the WHERE clause, plus the outbox event written from the same transaction
+ * only when a row was actually deleted.
+ *
+ * Closes codex F1 (detector race with heartbeat refresh): the staleBeforeMs cutoff
```
noe-charmet: Don't mention codex findings, they are not relevant for the reader (he does not have the tickets/review).

Focus on making the comments explain WHY this works as it is

libs/api/runners/src/db/jobs.ts
```diff
@@ -70,44 +70,157 @@ export async function claimJob(params: {
...
+    jobId: params.jobId,
+    runnerTokenId: params.runnerTokenId,
+    status: result.status,
+    output: result.output,
+    onMissing: 'throw',
+  });
+  // onMissing:'throw' guarantees a non-undefined return, but TS can't see that.
+  if (!finalized) throw new RunningJobNotFoundError(params.jobId);
+  return finalized;
+}
```
noe-charmet: This looks like `core` logic and not `db` logic you're never calling the db directly

libs/api/runners/src/db/jobs.ts
```diff
@@ -70,44 +70,157 @@ export async function claimJob(params: {
...
+      jobId: candidate.jobId,
+      staleBeforeMs: params.thresholdSeconds * 1000,
+      status: 'failed',
+      output: {reason: 'runner_disappeared'},
+      onMissing: 'noop',
+    });
+    if (finalized) failed += 1;
+  }
+  return {failed};
+}
```
noe-charmet: Move this to core and extract a db function to find stuck jobs

libs/api/runners/src/presentation/routes/heartbeat.ts
```diff
@@ -0,0 +1,51 @@
...
+ *                                                RETURNING cancellation_requested_at
+ *           ◄─{cancel}─                          ◄────────────────────
+ *
+ *   cancel:false → runner schedules next tick (single-flight setTimeout chain).
+ *   cancel:true  → runner aborts the per-job AbortController → kills the step's
+ *                  process group; the job will be reported as failed (or 404 if
+ *                  the orchestration timeout already finalized it).
+ *   404          → row is gone (orchestration timeout or stuck-job-detector won
+ *                  the race). Runner aborts and moves on.
+ */
```
noe-charmet: This is probably too detailed for an endpoint comment. We already detail it in other parets of the code maybe. And we need to focus on what's not explicit by code only and the WHY.

libs/api/runners/src/presentation/routes/heartbeat.ts
```diff
@@ -0,0 +1,51 @@
...
+export const heartbeatRoute = defineRoute({
+  method: 'POST',
+  path: '/:jobId/heartbeat',
+  description: 'Runner liveness ping; response carries the cancellation flag',
```
noe-charmet: This comment is for our external documentation it should be non-jargon / non-internal explanations in clear english for external developers

libs/api/workflows/src/presentation/subscribers/on-runner-job-completed.ts
```diff
@@ -3,12 +3,36 @@ import {logger} from '@shipfox/node-opentelemetry';
 import type {DomainEvent} from '@shipfox/node-outbox';
 import {temporalClient} from '@shipfox/node-temporal';
 
+/**
+ * Backend-determined terminal state wins (codex F8).
```
noe-charmet: Same here for codex findings

libs/api/workflows/src/presentation/subscribers/on-runner-job-completed.ts
```diff
@@ -3,12 +3,36 @@ import {logger} from '@shipfox/node-opentelemetry';
...
+ * If `jobOrchestration` already terminated via timeout (or via a prior runner
+ * completion), signaling the workflow handle throws `WorkflowNotFoundError` from
+ * `@temporalio/common`. We swallow it intentionally — a runner-reported success
+ * arriving after the workflow has declared the job failed cannot retroactively
+ * change the system-of-record state. The runner-side log of the run still shows
+ * what the runner thought, for forensics.
+ *
+ * Match by `error.name` rather than `instanceof` to keep this subscriber free of
+ * a direct `@temporalio/common` dependency and resilient across SDK minor versions.
+ */
```
noe-charmet: Do we need to explain this whole thing here ? Can we make things explicit in code to read ?

libs/api/workflows/src/temporal/activities/orchestration-activities.ts
```diff
@@ -130,3 +130,22 @@ export async function enqueueJobForRunner(params: {
     payload,
   });
 }
+
+/**
+ * Thin Temporal activity wrapper around the runners-module command. The runners
+ * module owns the SQL; this exists only so a workflow can call into it via
+ * Temporal's activity layer.
+ */
```
noe-charmet: Are the comments 100% needed/useful ?

libs/api/workflows/src/temporal/activities/orchestration-activities.ts
```diff
@@ -130,3 +130,22 @@ export async function enqueueJobForRunner(params: {
...
+export async function detectAndFailStuckJobsActivity(params: {
+  thresholdSeconds: number;
+}): Promise<{failed: number}> {
+  return await detectAndFailStuckJobs(params);
+}
+
+/**
+ * Thin wrapper. Keep the runtime cheap so the bounded retry policy in
+ * jobOrchestration can complete within `scheduleToCloseTimeout`.
+ */
```
noe-charmet: Are the comments 100% needed/useful ?

apps/runner/src/api-client.ts
noe-charmet: Done in f67539d. `requestJob`, `completeJob`, and `heartbeat` now return DTO types from `@shipfox/api-runners-dto` (`JobPayloadResponseDto`, `CompleteJobResponseDto`, `HeartbeatResponseDto`); `completeJob`'s body uses `CompleteJobBodyDto`.

apps/runner/src/config.ts
noe-charmet: Done in f67539d. `SHIPFOX_HEARTBEAT_MAX_STALE_MS` default is now 10_000.

apps/runner/src/heartbeat-loop.test.ts
noe-charmet: Good catch — added `await Promise.resolve()` before the `not.toHaveBeenCalled()` assertion in f67539d so a future regression that resolved synchronously would not slip past the immediate `expect`.

apps/runner/src/run-step.test.ts
noe-charmet: Done in f67539d — dropped the `(codex F9)` reference from the test name.

apps/runner/src/run-step.ts
noe-charmet: Done in f67539d. The comment now explains *why* `detached: true` is needed (Linux does not propagate signals down the parent chain, so without this the shell's grandchildren would survive a SIGKILL on just the immediate child) without referencing the review ticket.

apps/runner/src/runner.ts
noe-charmet: Done in f67539d — dropped the `Codex F5` marker; the comment now describes the actual behavior (404 from `/complete` is expected when the backend already finalized the job, must not be retried, must not fall into the outer execution-error path).

libs/api/runners/src/db/jobs.test.ts
noe-charmet: Done in f67539d. The runners-side `globalSetup` already truncates all four tables once, so I dropped the per-`beforeEach` TRUNCATE in the new `recordHeartbeat`, `requestJobCancellation`, and `detectAndFailStuckJobs` describes. Each test still creates a unique `workspaceId`, and the assertions in `detectAndFailStuckJobs` are now scoped (`runningJobsForTest()` filters by workspace, `outboxForJobs(jobIds)` filters by jobId) — no global-state assumptions remain.

libs/api/runners/src/db/jobs.ts
noe-charmet: Done in f67539d. Rewrote the `finalizeRunningJob` doc-comment to describe *why* folding the predicates into the DELETE makes the operation race-safe (heartbeat refresh can no longer cause a stale-snapshot finalize; two concurrent finalizers can't both observe the row and emit duplicate events) without referencing review tickets the reader does not have.

libs/api/runners/src/db/jobs.ts
noe-charmet: Done in f67539d. `completeJob` moved to `core/jobs.ts` (it composes `finalizeRunningJob` + the not-found throw, no direct DB access). `db/jobs.ts` now exports the primitives; `core/jobs.ts` exports the orchestration.

libs/api/runners/src/db/jobs.ts
noe-charmet: Done in f67539d. `detectAndFailStuckJobs` moved to `core/jobs.ts` and the iteration SELECT was extracted into a new `findStuckJobs` primitive in `db/jobs.ts`. Public surface from the package root is unchanged (`completeJob`, `detectAndFailStuckJobs` re-exported via `core/index.ts` → `src/index.ts`).

libs/api/runners/src/presentation/routes/heartbeat.ts
noe-charmet: Done in f67539d — removed the inline ASCII flow diagram from the route file. The flow lives in the plan and the heartbeat-loop module documentation; the route file now just declares the endpoint.

libs/api/runners/src/presentation/routes/heartbeat.ts
noe-charmet: Done in f67539d. New `description`: "Keeps a running job alive while the runner is still working on it. Returns whether the server has asked the runner to cancel."

libs/api/workflows/src/presentation/subscribers/on-runner-job-completed.ts
noe-charmet: Done in f67539d — dropped the `codex F8` reference. The subscriber now has a tight comment on *why* the late event is dropped (the workflow's status is authoritative; a runner result arriving after the timeout cannot retroactively change it).

libs/api/workflows/src/presentation/subscribers/on-runner-job-completed.ts
noe-charmet: Done in f67539d. Pulled the `WorkflowNotFoundError` matcher into a small `isWorkflowNotFound(err)` helper and trimmed the comment to one line on the policy. The control-flow is now self-explanatory in code (`if (isWorkflowNotFound(err)) { log.debug(...); return; } throw err;`).

libs/api/workflows/src/temporal/activities/orchestration-activities.ts
noe-charmet: Removed in f67539d — the wrapper body is the whole story; comment was redundant.

libs/api/workflows/src/temporal/activities/orchestration-activities.ts
noe-charmet: Removed in f67539d — same as above.

PR #96, opened 2026-05-31
Add local workflow dashboard integration

ThierryAbalea:
Add the local workflow integration path for the Foxlang V0 local service and expose it in the project dashboard.

The API now includes a dedicated local-workflows module and DTO package for service status, registered workflow metadata, fake alert triggering, run listing, and run detail retrieval. The dashboard consumes that contract to show local workflow registration, run history, run detail graphs/logs/source, and fake-alert execution from project pages.

This is intended to make the V0 local workflow loop reviewable from the platform UI while keeping the platform/API boundary HTTP-based and the Foxlang service isolated behind a small module-owned proxy.

Reviewers should pay particular attention to the new local-workflows contract shape and the run detail page state model, since those are the most likely surfaces to change as the V0 service evolves.

---
## Summary by cubic
Adds a new local-workflows API and dashboard UI to integrate the Foxlang V0 Local Service. You can view registered workflows, browse run history, inspect run details (graph, logs, source), and trigger fake alerts; docs now include a guided local demo and safe dashboard link tips.

- **New Features**
  - API: `@shipfox/api-local-workflows` exposes `/local-workflows/projects/:projectId` routes for status, workflows (list/get), runs (list/get), and fake-alerts (POST). DTOs and strict validation live in `@shipfox/api-local-workflows-dto` with mapped error codes.
  - Client: New local-workflows hooks, updated Project Runs/Workflows pages to use the local-service contract, and a Run Detail page with execution graph, logs, and source tabs.
  - Navigation/Docs/Tests: Adds `runs_.$runId` route, hides project tabs on run detail, surfaces run/status context in the nav bar, expands Foxlang V0 E2E setup in CONTRIBUTING, adds a “Local platform demo” with safe dashboard link guidance in AGENTS.md, and adds API/client tests.

- **Bug Fixes**
  - Preserves local-service path prefixes, rejects unsafe opaque IDs, and scopes generated run IDs to the project.
  - Fixes dashboard polling intervals and loading/empty fallback states.
  - Tightens DTOs and error handling to enforce the contract consistently.

<sup>Written for commit 4861916710e2b796d1a8927e3e16b7d8d24b33d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/platform/pull/96?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libs/api/local-workflows/src/core/local-workflows.ts
```diff
@@ -0,0 +1,310 @@
...
+  getWorkflow(workflowId: string): Promise<FoxlangWorkflowDetailResponseDto>;
+  listRuns(projectId?: string): Promise<FoxlangRunListResponseDto>;
+  getRun(runId: string, projectId?: string): Promise<FoxlangRunDetailResponseDto>;
+  triggerFakeAlert(
+    alert: Omit<FoxlangFakeMonitoringAlertRequestDto, 'run_id'>,
+    projectId?: string,
+  ): Promise<{run_id: string; result: FoxlangExecutionResponseDto}>;
+}
+
+export class LocalWorkflowsError extends Error {
```
noe-charmet, line 51: I understand the intent of avoiding duplication between layers here, and I agree that's generally a good goal.

That said, we haven't fully adopted the [error handling approach described in the Shipfox guidelines yet](https://github.com/ShipfoxHQ/shipfox/blob/main/CLAUDE.md#error-handling-node--fastify). At the `core` layer, our preference is to have fully branded business errors that describe the business condition itself and remain completely unaware of presentation-layer concerns.

A few reasons for this:

* It makes error handling more reliable when performing conditional checks, for example:

  ```ts
  if (error instanceof WorkflowNotFoundError) {
    // ...
  }
  ```
* Business-layer errors may be consumed by multiple upper layers. HTTP/presentation is one consumer, but the same business logic could also be used by async jobs, event handlers, CLI commands, etc., each with its own error-handling semantics.

Because of this, we generally prefer the presentation layer to explicitly translate business errors into presentation-specific errors. In the HTTP layer, for example, we use `ClientError` to represent user-facing HTTP errors, and our error handlers already know how to process those appropriately.

We also tend to favor explicit mappings at the endpoint level. While it may feel slightly more repetitive, it makes the translation decision visible and intentional. A given business error does not necessarily map to the same user-facing response in every context. For example, should `WorkflowNotFoundError` always produce the exact same HTTP response across all endpoints and methods? That may be true today, but it's not something we necessarily want to assume or encode globally.

For those reasons, I'd lean toward keeping the business error in the core layer and performing the translation explicitly in the presentation layer. We can also move the error handling guidelines over to this repo's `AGENTS.md`

ThierryAbalea:
Shipfox homepage is hosted on a private repository. So I am closing this PR

PR #99, opened 2026-06-01
Wire up changesets release flow and open libs for publishing

noe-charmet:
Sets up the changesets-based release flow end to end and makes every `libs/*` package publishable.

**Packages**

- Flip all 36 `libs/*` packages from `"private": true` to `"private": false`. The repo is open-source, so libs publish under `@shipfox/*` just like the existing `tools/*` packages.
- Remove the `changeset` / `version-packages` / `release` scripts from the workspace `package.json` — they're either redundant (`pnpm exec changeset` works directly) or now handled by the release workflow.

**Release workflow (`.github/workflows/release.yml`)**

- Manual dispatch only — pass an optional `package_name` filter, leave blank to publish everything eligible.
- Consume pending changesets → build (`turbo build type:emit`) → publish (`pnpm publish --recursive --provenance`) → open a release PR with the version bumps.
- npm auth via **trusted publishing** (OIDC, `id-token: write`) — no `NPM_TOKEN` in secrets.
- PR creation auth via a **GitHub App installation token** (`actions/create-github-app-token`) plus `sign-commits: true` on `peter-evans/create-pull-request` — no PAT and no GPG key in secrets. The App token is short-lived (~1h), scoped to one repo and two permissions.

**External setup needed before first run**

1. Configure trusted publishers on npmjs.com for each `@shipfox/*` package, pointing at `ShipfoxHQ/platform` + `release.yml`.
2. Create a GitHub App in the `ShipfoxHQ` org with `Contents: read & write` + `Pull requests: read & write`, install it on this repo only, and add:
   - Variable `RELEASE_BOT_APP_ID`
   - Secret `RELEASE_BOT_APP_PRIVATE_KEY`

**Docs and skills**

- Add a concise `Releases & Changesets` section to `CONTRIBUTING.md` with the when-rule, bump table, and a pointer to the workflow.
- Delete `DEVELOPMENT.md` (duplicated `CONTRIBUTING.md` and referenced the commercial repo).
- Trim `.changeset/README.md` to a 4-line pointer.
- Rewrite `.agents/skills/generate-changeset/SKILL.md` for the agent path (no interactive CLI), reference `CONTRIBUTING.md` for the when-rule, drop the patch-only rule, add a worked example, declare `allowed-tools`.
- Update `.agents/skills/pr/SKILL.md` with a "Step 3: Add a Changeset If Required" that points to the same single source of truth.

Notes for review:

- `@shipfox/api-workflows` tests are flaky on a stale local Postgres (pre-existing on `main`, unrelated to this diff).
- The release workflow won't actually run until the npm trusted publishers and the GitHub App are configured — it's safe to merge first.


noe-charmet:
Addressed cubic-dev-ai's review in 2b97452:

**Missing `version` fields (11 findings)** — added `"version": "0.0.0"` to every `libs/*` package flipped to public. Both `pnpm publish` and `changeset version` need a version baseline; existing changesets will bump from there on first release.

**Workflow `Bump versions` step (P1)** — `changeset version` now only runs on the all-packages path. The single-package path skips it entirely, so we no longer consume pending changesets that won't be published.

**Workflow `Publish a specific package` step (P2)** — `inputs.package_name` now flows through an `env:` block and is referenced as a quoted shell variable, removing the direct `${{ ... }}` interpolation.

Also swapped the all-packages publish step from `pnpm publish --recursive` to `pnpm exec changeset publish` so the workflow only releases packages whose version was actually bumped by `changeset version` — matches the new safer model.

PR #123, opened 2026-06-05
Formalize workflow runtime semantics

ThierryAbalea:
Formalizes the workflow runtime in reviewable layers: a shared workflow-language package, a normalized WorkflowIR, reusable static semantics, run creation from IR, a pure job-level runtime kernel, and Temporal orchestration that delegates scheduling decisions to that kernel.

This PR keeps the existing YAML and HTTP wire surface compatible for PR1 while moving runtime behavior behind explicit, tested concepts. The goal is to make future workflow-language changes start from a documented requirement, normalize through IR, validate through shared semantics, and only then reach persistence, Temporal, runners, or triggers. Repo-owned generated docs under `docs/formalizing-shipfox-runtime/` capture that process, the included PR1 scope, and the deliberately deferred work.

The branch is intentionally split into thirteen commits so it can be reviewed commit-by-commit: twelve implementation commits followed by one consolidated generated-documentation commit. Each implementation or documentation slice had a Claude consult before commit; agreed recommendations were either implemented in the relevant slice or recorded as deferred follow-up.

Reviewers should pay closest attention to the boundaries between the new `@shipfox/api-workflow-language` package, `@shipfox/api-definitions`, and `@shipfox/api-workflows`; the WorkflowIR normalization and static-semantics diagnostics; and the runtime command interpreter in Temporal. The current durable execution host preserves the pre-existing batch behavior for sibling child workflows, and pipelined scheduling is documented as deferred.

The formalization docs now include generated references for the YAML surface, CUE formalization artifact, expression boundary, core IR, diagnostics, runtime transitions, golden traces, Temporal durable execution host adapters, and the process for adding future concepts. `010-future-platform-use-cases.md` remains hand-authored and exploratory; its YAML snippets are explicitly not accepted PR1 syntax.

No changeset is included because the touched workspace packages are private and this does not change a publishable package API.

Validated with affected package graph `check`, `type`, and `test`; generated-doc determinism checks; a forbidden temporary-path scan over committed repo files; and broad affected `build`, `test`, and `check` against `origin/main`.

---
## Summary by cubic
Formalizes the workflow runtime around a shared language and normalized IR, with a pure job-level transition kernel that Temporal now delegates to. Adds stricter static validation, deterministic formalization docs, and tighter dependency rules.

- New Features
  - Added `@shipfox/api-workflow-language`: YAML surface schemas/parser, CUE schema artifact, WorkflowIR normalizer/ID helpers, static diagnostics/validator, and a typed default run-step acceptance policy; includes a deterministic doc generator under `docs/formalizing-shipfox-runtime/`.
  - Introduced a pure runtime kernel (`runtime-state`, `runtime-event`, `runtime-command`, `transition`) with targeted tests and golden traces.
  - Run creation now builds `WorkflowIR` from definitions and validates static semantics; invalid definitions raise `InvalidWorkflowDefinitionError`. Static semantics validate dependency edge targets (`SS004_UNKNOWN_DEPENDENT_JOB`) and limit cycle diagnostics to only cyclic jobs.
  - Temporal orchestration interprets kernel commands for start/cancel/complete while preserving existing sibling batch behavior.
  - Generated docs expanded (surface schema, IR types, static diagnostics, runtime transitions, durable execution host, surface language portability, WorkflowIR rationale, and an architecture overview) and are now produced by generators in `@shipfox/api-workflow-language` and `@shipfox/api-workflows`.

- Refactors
  - Definitions and DTOs re-export surface schemas from `@shipfox/api-workflow-language`; replaced local YAML/DAG validation with shared parsing and static checks; removed `validate-dag`.
  - Workflow run persistence accepts `WorkflowIR` (normalized steps, edges, runners) instead of raw spec; queries and tests updated.
  - Dependency rules now scope to the current package and add two guards: `@shipfox/api-workflow-language` has no feature/runtime deps, and `@shipfox/*-dto` packages may only depend on other `*-dto` packages at runtime. No changes to public wire contracts or publishable package APIs.

<sup>Written for commit 15b0d1c2601a062e121b87ca67ab9dd4b903f2f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


.dependency-cruiser.cjs
```diff
@@ -1,19 +1,41 @@
...
+  comment:
+    '@shipfox/api-workflow-language is a lower-level language package and must not import feature, runtime, adapter, or app packages.',
+  severity: 'error',
+  from: {
+    path: '^(src|test|scripts)/',
+  },
+  to: {
+    path: '^\\.\\./(?:definitions|workflows|triggers|runners)(?:-dto)?/(?:src|dist)/|^(?:\\.\\./)+client/|^(?:\\.\\./)+apps/|^(?:\\.\\./)+shared/node/(?:drizzle|fastify|temporal)/(?:src|dist)/|^node_modules/(?:drizzle-orm|fastify|@temporalio/)',
+  },
+};
```
noe-charmet, line 30: I totally agree that this language definition package should not import anything, however the issue here is not that we need a rule specifically for this package.
The package itself is misplaced as `libs/api/...` as it is not in any ways a business-logic / component module of the `api` application. It rather describes a utility like library which is in itself not specific to a technology (ie: it does not expose specific APIs which require node, react or some other framework). It therefore should live somewhere under `libs/shared/common`. Being there will clearly signify that it is a shared utility library, it can be imported from nearly anywhere, however it itself can only import other shared utility libraries.

libs/api/workflows/package.json
noe-charmet, line 1: Could we keep this PR lighter and make things easier to review by scoping it down on only working on the language reference package, and not on its usage or adapting it to the `api` application in any ways ?
It's currently touching both the definitions and workflow modules in many places, which polutes the original intent and potential discussions on the changes this PR aim to bring.

libs/api/workflow-language/src/core/expressions/index.ts
noe-charmet, line 1: Do we need this file here ?

libs/api/workflow-language/src/dependency-boundary.test.ts
noe-charmet, line 1: I don't believe having tests on the dependency rules package here is really something we want. It's mostly a symptom of this package not being placed correctly it seems, and therefore needing custom overrides.
Overall it sounds like having tests to test the linter or the test engine itself. Not something that's really realted to this specific package.

docs/formalizing-shipfox-runtime/README.md
noe-charmet, line 1: Overall, I feel like this documentation reads as fairly "LLM-generated" and could be more accessible. The goal should be for it to be readable by anyone, regardless of their English skills or their computer science background.

A couple of specifics:
- Some concepts are tied to the current local setup of your LLM sessions, like PR1. It's already hard to tell what that refers to today, so a reader coming to this fresh in a few months will have an even harder time.
- The tone leans quite academic, closer to a research paper, and leans on a lot of technical jargon that I think we can dial back.

That said, this is a great initiative. Having a proper specification of the configuration capabilities is really valuable. I'd just like us to figure out how to make it lighter and more approachable, both for us and for our future team and community.

docs/formalizing-shipfox-runtime/README.md
```diff
@@ -0,0 +1,90 @@
...
+- It keeps Temporal, persistence, outbox, runner, and future integration effects outside the pure kernel.
+- It gives future concepts a required path through requirements, surface, IR, diagnostics, runtime semantics, durable execution host changes, generated docs, and tests.
+
+## Status Meanings
+
+| Status | Meaning |
+| --- | --- |
+| `normative` | Describes committed PR1 behavior or the committed process for extending it. |
+| `deferred` | Documents a concept that PR1 intentionally does not implement. |
+| `exploratory` | Captures future product or platform examples. Exploratory YAML snippets are not accepted PR1 syntax unless another normative document says so. |
```
noe-charmet, line 92: I feel like the specs in the repository (and anywhere in the codebase, in fact) aren't really the right place to store future evolution or "to be decided" ideas.
A couple of reasons:

It's how we end up with all those `// TODO: notes` that never actually get done.
Changes or proposals to the spec then end up mixing two discussions: the current proposal (the PR itself) and future work or ideas.

Overall, it seems like we'd be better off splitting work to be done (tickets/issues) from the specifications and code.

docs/formalizing-shipfox-runtime/002-cue-schema.md
noe-charmet, line 1: CUE is great for schema definition, but I don't think it's the best fit for us. I'd vouch for keeping JSON Schema / OpenAPI as the source of truth.

For pure schema definition, the two are very close. Both handle types, nesting, enums, required/optional fields, and pattern/range constraints. CUE does add more advanced capabilities on top: config unification/layering, real expressions and computation, first-class defaults, and cross-field constraints.

The thing is, most of those extras go beyond what a config schema actually needs to do for us, which is to validate and document the YAML. On the fundamentals that matter for our usage and for driving adoption, JSON Schema wins:

- It's ubiquitous and has mature libraries in every language. We can even pass it as-is to our HTTP endpoints config and have them validate against it.
- Validating against CUE would require the cue CLI, since there's no strong JS/TS library. That adds a layer of complexity and another place for things to go wrong.
- CUE's compatibility only goes one way (JSON Schema to CUE), not the reverse. We can validate CUE config files against a JSON Schema, but we can't emit a JSON Schema out of a CUE schema.

This leads me to think we'd be better off with a package that exposes the language spec as a JSON Schema. With that approach:
- We can validate natively in any language.
- Reusing the schema across the JS/TS ecosystem is extremely easy.
- We can support JSON/YAML/TOML/CUE parsing and validation with very minimal effort.

docs/formalizing-shipfox-runtime/004-core-ir.md
noe-charmet, line 1: I'm worried that centralizing everything in a single "language" module with formal specifications goes against the DDD/modularity principles we've been following.

If we step back and think about ownership, I'd argue we actually have three distinct layers, each with different needs:
1. The external representation (the user's YAML config). This is the one place where we genuinely need clear, public specs and documentation. A shared package with a well-defined JSON Schema feels like the easiest foundation, since it can serve definition, validation, and documentation all at once.
2. Parsing the external representation into our internal format. This is the job of the definitions module. It exposes some public interfaces within the monorepo (including how it represents workflow definitions), but I think that representation should remain solely owned by this module. We need tests to confirm parsing produces what we expect, but I don't think it warrants a formal specification, and its internal objects shouldn't live outside its ownership boundary.
3. The orchestration layers downstream (workflow/job management, execution on the runner). These consume DTO-like objects from definitions and each maintain their own internal representation suited to their specific goals. I'd push for letting each module own its entities in a pure DDD way (a Workflow isn't the same thing across modules), backed by tests that assert each layer's behavior as the source of truth.

The key distinction for me: we're not building a programming language for the general public, we're defining a configuration format for the Shipfox apps. So:

- Formalizing and documenting the external representation is very important.
- Over-formalizing the internals and their exact behavior feels like overkill.

This also seems to be the path most comparable tools take. Kubernetes, ArgoCD, and GitLab CI all cleanly define their JSON Schema / OpenAPI specs, but none of them take on the complexity and weight of full formal specifications.

ThierryAbalea:
this PR was replaced by a bunch of PRs already merged

PR #125, opened 2026-06-08
Add workflow document contract

ThierryAbalea:
Adds `@shipfox/workflow-document` under `libs/shared/workflow/document` as the external Shipfox workflow document contract.

This is the first PR in the rewritten workflow formalization stack. It is deliberately limited to the authoring-facing document shape and does not wire definitions, persistence, trigger matching, runtime scheduling, Temporal orchestration, or gate behavior.

The package exports:

- the top-level Zod schema and inferred TypeScript types for `WorkflowDocument`;
- `parseWorkflowDocument(input)`, which returns a typed document or throws;
- `InvalidWorkflowDocumentError`, a typed `Error` subclass with the original Zod error available through `cause` and `validationError`;
- a small `simple build` example validated by tests.

This PR also removes the previous numeric workflow-document diagnostic direction. For JavaScript APIs, invalid documents now use typed errors and readable kebab-case codes only where they are useful to presentation or serialized boundaries. `AGENTS.md` is updated with that repo-wide error-management guideline.

JSON Schema and generated workflow docs are intentionally out of scope for this stack rewrite. They can come back later when there is a concrete consumer.

## Validation

- `mise exec -- turbo test --filter=@shipfox/workflow-document`
- `mise exec -- turbo type --filter=@shipfox/workflow-document`
- `mise exec -- turbo check --filter=@shipfox/workflow-document`
- `node .claude/skills/readme-writer/scripts/readability.mjs libs/shared/workflow/document/README.md`

## Stack

Part 1 of 7.

Depends on: none
Next: #126

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #125 | `workflow-config-reference` | `main` |
| 2 | #126 | `workflow-yaml-surface` | `workflow-config-reference` |
| 3 | #127 | `workflow-expressions` | `workflow-yaml-surface` |
| 4 | #128 | `workflow-model-ir` | `workflow-expressions` |
| 5 | #129 | `workflow-definitions-integration` | `workflow-model-ir` |
| 6 | #130 | `workflow-runtime-scheduling` | `workflow-definitions-integration` |
| 7 | #131 | `workflow-runtime-kernel-spike` | `workflow-runtime-scheduling` |


libs/shared/workflow/document/package.json
noe-charmet, line 1: Unless you have some other specific packages in mind to add to the `libs/shared/workflow` dir, I would suggest being YAGNI and simplifying to `libs/shared/workflow-document`. The path already conventionally explains that it is a library which do not depend on any specific technology (ie: not react or node specific), and that ti's named `workflow-document`

libs/shared/workflow/document/src/document/workflow-document.ts
```diff
@@ -0,0 +1,51 @@
...
+  source: z.string().min(1),
+  with: z.record(z.string(), z.unknown()).optional(),
+  filter: z.string().min(1).optional(),
+} satisfies z.ZodRawShape;
+
+export const workflowDocumentTriggerSchema = z.union([
+  z.strictObject({
+    ...workflowDocumentTriggerBaseSchema,
+    event: z.string().min(1),
+    on: z.never().optional(),
```
noe-charmet: I'm not sure of the need to express this type in such a way using `never` types. They probably can be removed entirely IMO. It looks like what you want to express would be a TS type like this one.

```typescript
type Trigger = BaseTrigger & (
  {
    event: string
  }
  | {
    on: string | string[]
  }
)
```

However, we could discuss simplifying to only one key unless we have a good usage for this case.

noe-charmet:
Instead of having both a JSON schema and a Zod schema and having to keep both in sync here, I'd suggest choosing one as the source of truth.

Zod is convenient as it also allows easily deriving TS types from it. Zod exposes a method to export to [JSON-schema](https://zod.dev/json-schema). We could have it generated as part of this package's build step and git-ignored if we want to expose a JSON schema to external users through this package.

libs/shared/workflow/document/src/document/workflow-document.ts
```diff
@@ -0,0 +1,51 @@
...
+  source: z.string().min(1),
+  with: z.record(z.string(), z.unknown()).optional(),
+  filter: z.string().min(1).optional(),
+} satisfies z.ZodRawShape;
+
+export const workflowConfigTriggerSchema = z.union([
+  z.strictObject({
+    ...workflowConfigTriggerBaseSchema,
+    event: z.string().min(1),
+    on: z.never().optional(),
```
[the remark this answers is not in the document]
ThierryAbalea: remove "on", "event" is the field to use instead

libs/shared/workflow/document/package.json
ThierryAbalea: 2 other modules are planned:
- `@shipfox/workflow-model`: Pure semantic model layer that normalizes `WorkflowDocument`, applies defaults/shorthands, builds graph edges, and reports static diagnostics.

- `@shipfox/workflow-runtime`: Pure runtime decision layer that evaluates a normalized workflow/run plan and decides what can run, wait, fail, skip, or transition next.

ThierryAbalea:
> Instead of having both a JSON schema and a Zod schema and having to keep both in sync here, I'd suggest choosing one as the source of truth.
> 
> Zod is convenient as it also allows easily deriving TS types from it. Zod exposes a method to export to [JSON-schema](https://zod.dev/json-schema). We could have it generated as part of this package's build step and git-ignored if we want to expose a JSON schema to external users through this package.

I removed the JSON schema following our sync conversation. We will reintroduce it properly when needed.

libs/shared/workflow/document/src/document/workflow-document.ts
ThierryAbalea: i removed "on". "event" is the field to use instead

ThierryAbalea:
I added stable errors/diagnostics to `@shipfox/workflow-document`, then trimmed the raw Zod tests.

The split is now:

- Zod tests protect the public document shape exposed by `workflowDocumentSchema`.
- Diagnostic tests protect the stable `WFD` codes, paths, and messages exposed by `validateWorkflowDocument`.

The diagnostic scope here is only document-shape errors: missing fields, wrong types, empty arrays/maps, unknown fields, and unsupported fields like trigger `on`.

Semantic errors such as unknown `needs`, cycles, runner resolution, or invalid expression references should live later in `workflow-model`. Runtime-state errors such as invalid transitions, duplicate completions, or signals for non-waiting steps should live later in `workflow-runtime`.

As a reminder here are the scope of the future modules:
- @shipfox/workflow-model: Pure semantic model layer that normalizes WorkflowDocument, applies defaults/shorthands, builds graph edges, and reports static diagnostics.
- @shipfox/workflow-runtime: Pure runtime decision layer that evaluates a normalized workflow/run plan and decides what can run, wait, fail, skip, or transition next.

libs/shared/workflow/document/src/document/workflow-document-diagnostics.ts
```diff
@@ -0,0 +1,255 @@
...
+  | 'WFD002'
+  | 'WFD003'
+  | 'WFD004'
+  | 'WFD005'
+  | 'WFD101'
+  | 'WFD102'
+  | 'WFD201'
+  | 'WFD202'
+  | 'WFD301'
+  | 'WFD302';
```
noe-charmet: Image

Do we plan on adding a phone support line or print a book with all our error codes and their counter-measures ?

In all seriousness, these are not really great for DevX 😆
The need to have computer readable errors is quite understandable, especially if this can be used programmatically or even in tests.
However we can do a lot better than random number error codes.

For non-JS APIs (understand that get serialized) such as over HTTP, we always represent error codes with a kebab case string `invalid-user-identity`. This makes it readable for both the implementer and the user. A lot easier to read a test that asserts `error.code = "invalid-schema"`.

For JS APIs we can use the `Error` class inheritance and make branded errors from it. This is kind of what you're trying to mimic with your interface it seems.

```typescript
class MissingSudoKeyError extends Error {
  constructor(path: string) {
    super(`Missing sudo key at path ${path}`)
  }
}

// In the part using it
if (!('sudo' in object)) throw new MissingSudoKeyError(path);

// In the test
expect(error).toBeInstanceOf(MissingSudoKeyError)
```

libs/shared/workflow/document/src/document/workflow-document-diagnostics.ts
```diff
@@ -0,0 +1,255 @@
...
+  const path = [...toDiagnosticPath(basePath), key];
+
+  if (isTriggerPath(basePath) && key === 'on') {
+    return diagnostic({
+      code: 'WFD101',
+      message: 'Trigger field "on" is not supported; use "event".',
+      path,
+      details: {field: key},
+    });
+  }
```
noe-charmet: Why do we need such a specific over a key that was never part of the spec ?

libs/shared/workflow/document/src/document/workflow-document-diagnostics.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  return {
+    valid: false,
+    diagnostics: result.error.issues.flatMap((issue) =>
+      toWorkflowDocumentDiagnostics(issue, result.error.issues, input),
+    ),
+  };
+}
+
+function toWorkflowDocumentDiagnostics(
```
noe-charmet: Overall I feel like we don't really need to do all this wrapping.
- Base errors are already somewhat user-friendly
- This will only increase in complexity as the schema grows

If we want to keep a wrapping for some reason, I would only re-wrap things into a custom error class, but not attempt any smart re-mapping or re-creation of error messages. IMO we should focus on defining the schema as to have it be explicit with its errors when parsing errors happen.

ThierryAbalea:
Stack rewrite completed after the module-organization review. The active stack is now 7 PRs (#125-#131). The old post-#125 commits and obsolete #132 should be considered superseded by the rewritten stack.

AGENTS.md
noe-charmet, line 1: I have added [this PR](https://github.com/ShipfoxHQ/shipfox/pull/141) with extensive documentation to `AGENTS.md`, maybe that these changes should be removed from the scope of this PR ?

libs/shared/workflow/document/src/examples/simple-build.ts
noe-charmet, line 1: Not a blocker, however we tend to place "test" files under `/test/data` or `/test/fixtures`. This allows keeping your test nearby the tested function, while keeping all the example payloads/strutures nicely outside of the `src` directory.
The package and vite config are already setup with aliases to pick-up from `#test/*`.

In this situation you could add your tests to `libs/shared/workflow/document/src/document/workflow-document-parser.test.ts`, while referencing JSON files in `/test/data/...`

libs/shared/workflow/document/src/document/workflow-document.ts
```diff
@@ -0,0 +1,43 @@
+import {z} from 'zod';
+
+const stringOrStringArraySchema = z.union([z.string().min(1), z.array(z.string().min(1)).min(1)]);
+const nonEmptyRecordSchema = <ValueSchema extends z.ZodType>(valueSchema: ValueSchema) =>
+  z
+    .record(z.string().min(1), valueSchema)
+    .refine((value) => Object.keys(value).length > 0, {message: 'Expected at least one entry'});
+
+// This package owns the syntactic external document shape. Definitions-layer
+// semantic rules, defaults, and internal representations stay in definitions.
```
noe-charmet: Very LLMish comment here 😄
Seems like README.md already covers this

noe-charmet (dismissed):
Good job, this lands with a lot simpler and straight-forward code 💯 
Left a few comments which address non-content issues you are welcome to look at before merging.

AGENTS.md
ThierryAbalea: Done. I removed the AGENTS.md change from PR #125 and restacked/pushed the stack, so AGENTS.md is no longer part of this PR diff.

libs/shared/workflow/document/src/examples/simple-build.ts
ThierryAbalea: Done. The simple-build payload now lives outside src at libs/shared/workflow/document/test/data/simple-build.json, and the validation coverage is in src/document/workflow-document-parser.test.ts. I restacked/pushed the stack, and the old src/examples file is no longer in the PR diff.

noe-charmet (dismissed):

libs/shared/workflow/document/src/document/workflow-document.ts
ThierryAbalea: Done. I removed the redundant inline comment; the package role stays documented in README.md instead.

noe-charmet (approved):

PR #126, opened 2026-06-08
Add workflow expression modules

ThierryAbalea:
## Objective

Adds Shipfox-owned expression modules that hide the selected CEL implementation behind stable package APIs.

This PR intentionally keeps expression parsing and evaluation separate from workflow parsing/modeling. It establishes the expression contract that later workflow-model work can depend on without exposing `@gresb/cel-javascript` directly.

## Scope

- Adds `@shipfox/expression-language` in `libs/shared/expression/language`.
  - Parses and type-checks CEL source with a caller-provided type environment.
  - Returns `WorkflowExpression` values with a branded `ValidCelExpression` source.
  - Documents the lexical workflow expression-context model.
- Adds `@shipfox/expression-evaluator` in `libs/shared/expression/evaluator`.
  - Evaluates validated expressions against caller-provided runtime values.
  - Keeps evaluation deterministic and side-effect-free.
- Patches `@gresb/cel-javascript@0.1.1` packaging so it can be imported from Shipfox ESM packages.

Persisted expressions remain only:

```ts
{ language: 'cel', source: string }
```

No CEL AST, checked protobuf, compiled object, or vendor metadata is persisted. Trigger-filter expression parsing/type-checking is deferred until event-schema-backed type environments are introduced.

## Validation

- `mise exec -- turbo build --filter=@shipfox/expression-language --filter=@shipfox/expression-evaluator`
- `mise exec -- turbo type --filter=@shipfox/expression-language --filter=@shipfox/expression-evaluator`
- `mise exec -- turbo test --filter=@shipfox/expression-language --filter=@shipfox/expression-evaluator`
- `mise exec -- turbo check --filter=@shipfox/expression-language --filter=@shipfox/expression-evaluator`
- README readability checks for both expression packages

## Review

Claude consult review was run for this rewritten PR. Accepted recommendations were applied, including broader scalar type coverage, less brittle vendor-error assertions, extra evaluator tests, and README notes about the temporary vendor patch.

## Stack

Part 2 of 7.

Depends on: #125
Next: #127

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 | #125 | `workflow-config-reference` | `main` |
| 2 (current) | #126 | `workflow-yaml-surface` | `workflow-config-reference` |
| 3 | #127 | `workflow-expressions` | `workflow-yaml-surface` |
| 4 | #128 | `workflow-model-ir` | `workflow-expressions` |
| 5 | #129 | `workflow-definitions-integration` | `workflow-model-ir` |
| 6 | #130 | `workflow-runtime-scheduling` | `workflow-definitions-integration` |
| 7 | #131 | `workflow-runtime-kernel-spike` | `workflow-runtime-scheduling` |

### DIFF
--- a/.changeset/calm-lemons-express.md
+++ b/.changeset/calm-lemons-express.md
@@ -1,0 +1,6 @@
+---
+"@shipfox/expression-language": minor
+"@shipfox/expression-evaluator": minor
+---
+
+Adds shared CEL expression language and evaluator packages behind Shipfox-owned APIs.
--- a/libs/shared/expression/evaluator/README.md
+++ b/libs/shared/expression/evaluator/README.md
@@ -1,0 +1,67 @@
+# Expression Evaluator
+
+Shared CEL evaluation for validated Shipfox expressions.
+
+## What it does
+
+- `evaluateWorkflowExpression` runs a validated `WorkflowExpression` against supplied values.
+- `evaluateWorkflowPredicate` returns `true` only when the expression result is the boolean `true`.
+- `WorkflowExpressionEvaluationError` wraps evaluation failures from the CEL engine.
+
+Use this package after an expression has already been accepted by `@shipfox/expression-language`. The caller supplies all values. The evaluator does not fetch data.
+
+Think of this as the last step. The text was checked before. Now the caller gives the values for this run, and this part returns a value. It should be easy to test because it only uses what the caller gives it.
+
+## Installation
+
+```sh
+pnpm add @shipfox/expression-evaluator
+```
+
+## Usage
+
+```ts
+import {createWorkflowExpression} from '@shipfox/expression-language';
+import {evaluateWorkflowPredicate} from '@shipfox/expression-evaluator';
+
+const expression = createWorkflowExpression({
+  source: 'event.conclusion == "success"',
+  typeEnvironment: {
+    event: {kind: 'object', fields: {conclusion: 'string'}},
+  },
+});
+
+const passed = evaluateWorkflowPredicate(expression, {
+  event: {conclusion: 'success'},
+});
+```
+
+## Behavior Notes
+
+- Evaluation is deterministic and side-effect-free.
+- The caller must pass values that match the type-checked context.
+- The evaluator does not read secrets, database rows, events, files, or external services.
+- The CEL engine stays hidden behind this package.
+
+Do not put side effects here. Do not read state here. Code that needs a row, file, secret, or event must get it before calling this package.
+
+When you add a new call site, build the data first. Then pass that data in one plain object. A test should be able to pass the same object and get the same answer.
+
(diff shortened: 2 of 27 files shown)


libs/shared/workflow/yaml/src/parser/parse-workflow-yaml.ts
```diff
@@ -0,0 +1,111 @@
...
+export type WorkflowYamlDiagnosticSeverity = 'error';
+export type WorkflowYamlDiagnosticPathSegment = WorkflowDocumentDiagnosticPathSegment;
+
+export interface WorkflowYamlDiagnostic {
+  code: WorkflowYamlDiagnosticCode;
+  severity: WorkflowYamlDiagnosticSeverity;
+  message: string;
+  path: readonly WorkflowYamlDiagnosticPathSegment[];
+  details?: Readonly<Record<string, unknown>>;
+}
```
noe-charmet: Really not a huge fan of this way of managing errors.

- We prefer `core` layer branded `Error` classes (ex: `libs/api/definitions/src/core/errors.ts`
- We are OK with the imperative `try/catch` mode, try to force everything into a functional return type system creates another pattern which becomes noisy in the code.

libs/shared/workflow/yaml/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@shipfox/workflow-yaml",
```
noe-charmet: I don't think that we need another package specifically for using the schema from YAML.

- If we plan on having this re-used anywhere, it can probably be directly embedded within the existing `workflow-document` package.
- If we don't plan on having it anywhere else than the API application, it seems like `api-definition` package would be best.

noe-charmet:
This seems to overlap entirely with what we do in `/libs/api/definitions/src/core/parse-definition.ts`.
I would suggest reviewing your plan to figure out how to start making this work into the current flow instead of building a completely new branch.
It looks like we need to bring the new schema inside of the current logic first thing.

noe-charmet:
Could we add a mode to allow non-strict value checking ?
If we want to validate that something is a valid CEL expression, while however not knowing yet exactly which pieces of context will be passed to it. For integration events typically we will have very little ways of knowing and documenting exactly all payload shapes and use them to validate at definition read time.

noe-charmet:
Do we actually need two packages here ? Aren't the scopes close enough that it would make sense to keep them in a single one ? Do we have specific cases in mind of consumers only needing one of both ?

ThierryAbalea:
Addressed the Cubic findings: removed the source-only package import map and switched the package entrypoint to a relative export, and removed the unsupported `uint` scalar from the expression type vocabulary instead of mapping it to CEL `int`. Local affected checks passed after the change.

PR #127, opened 2026-06-08
Add definitions workflow YAML parser

ThierryAbalea:
## Objective

Adds the definitions-owned workflow YAML parser.

This replaces the old standalone workflow YAML package approach. YAML is now treated as an ingestion concern for `@shipfox/api-definitions`, while the accepted document shape stays owned by `@shipfox/workflow-document`.

## Scope

- Adds `libs/api/definitions/src/core/workflow-yaml`.
  - `parseWorkflowYaml(source)` parses YAML source text into a `WorkflowDocument`.
  - `InvalidWorkflowYamlError` reports YAML syntax errors and non-object roots.
  - Parsed objects are delegated to `parseWorkflowDocument` from `@shipfox/workflow-document`.
- Adds golden fixtures for a valid `simple-build` YAML document.
- Adds invalid fixtures for YAML syntax and document-shape validation.
- Adds a module README written with the Shipfox README guidance.
- Adds the `@shipfox/workflow-document` workspace dependency to `@shipfox/api-definitions`.

Document validation errors intentionally propagate as `InvalidWorkflowDocumentError`. The YAML module owns YAML parsing errors only; later definitions integration can map both typed errors at the route or sync boundary.

## Validation

Passed:

- `mise exec -- turbo build --filter=@shipfox/api-definitions`
- `mise exec -- turbo type --filter=@shipfox/api-definitions`
- `mise exec -- turbo check --filter=@shipfox/api-definitions`
- `node .claude/skills/readme-writer/scripts/readability.mjs libs/api/definitions/src/core/workflow-yaml/README.md`
- Pure parser Vitest run with a temporary DB-free config: 1 file, 6 tests passed

Blocked:

- `mise exec -- turbo test --filter=@shipfox/api-definitions`

Both are blocked because the `@shipfox/api-definitions` package-level Vitest global setup requires Postgres and Docker is not running in this environment. `docker compose up -d` failed because the Docker daemon socket is missing at `/Users/thierry.abalea/.docker/run/docker.sock`.

## Review

Claude consult review was run for this rewritten PR. Accepted recommendations were applied:

- loosened YAML syntax location assertions so tests do not depend on exact js-yaml EOF marks;
- kept document validation errors unwrapped;
- recorded the no-DB test lane as a deferred follow-up outside the repo planning docs.

## Stack

Part 1 of 5.

Depends on: none
Next: #128

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #127 | `workflow-expressions` | `main` |
| 2 | #128 | `workflow-model-ir` | `workflow-expressions` |
| 3 | #129 | `workflow-definitions-integration` | `workflow-model-ir` |
| 4 | #130 | `workflow-runtime-scheduling` | `workflow-definitions-integration` |
| 5 | #131 | `workflow-runtime-kernel-spike` | `workflow-runtime-scheduling` |


noe-charmet:
I have some doubts about whether we should be building our own expression language. We can dig into the details together, but my personal preference would be to adopt an existing solution for expression evaluation rather than build and maintain our own, unless we identify specific needs that nothing out there can handle.

I see two main approaches:
- String based expressions, like [Google's CEL](https://github.com/marcbachmann/cel-js) (or [Filtrex](https://github.com/joewalnes/filtrex) in JS).
- Operators expressed directly in JSON/YAML, which is what [JSONLogic](https://jsonlogic.com/operations.html#var) does. This can make things easier for users to reason about, with GitHub Actions' expression syntax being a good example of where that readability tends to break down.

Both approaches natively take a context object for variable interpolation. I think we also need to start thinking about what our context should contain, based on where the evaluation is called from. That raises a question of ownership too: where does this live, and what's the runtime path for it?

At a minimum, we need to run evaluations in two places: the workflow module (if conditions on jobs) and the trigger module (trigger filtering). So the question becomes: should this be an external shared library that both consume, or part of the workflow module exposing an external API (which would let us centralize the context, typically)?

libs/api/definitions/test/fixtures/workflow-yaml/invalid/invalid-yaml-syntax.yaml
noe-charmet, line 1: You can move these specifically to a `/test/fixtures...` directory outside of `src` that we designed for this purpose. Helps keep only the actual code files and their directly related test side-by-side in the `src` directory.
Vite and the package manage know how to link to it with `#test/...`.

noe-charmet (dismissed):

libs/api/definitions/test/fixtures/workflow-yaml/invalid/invalid-yaml-syntax.yaml
ThierryAbalea: Done. The workflow YAML fixtures are now outside src under libs/api/definitions/test/fixtures/workflow-yaml/..., and parse-workflow-yaml.test.ts reads them through the test fixture path. I restacked/pushed the stack so the PR diff reflects the new layout.

PR #128, opened 2026-06-08
Add definitions workflow model normalization

ThierryAbalea:
## Objective

Adds the definitions-owned `WorkflowModel` layer.

`WorkflowModel` is the semantic model produced after a workflow document has passed the shared document contract. It is useful now with YAML as the only authoring surface because it gives downstream code one normalized graph shape for jobs, steps, runners, ids, and dependencies.

## Scope

- Adds `libs/api/definitions/src/core/entities/workflow-model.ts`.
- Adds `libs/api/definitions/src/core/workflow-model`.
  - `normalizeWorkflowDocument(document)` returns `WorkflowModel`.
  - `InvalidWorkflowModelError` reports semantic issues with kebab-case issue codes.
- Renames the existing platform entity file:
  - `src/core/entities/definition.ts` -> `src/core/entities/workflow-definition.ts`
  - Existing `WorkflowSpec` remains for legacy persistence/integration until PR #129.
- Adds tests for:
  - stable ids and edge cases;
  - top-level runner defaults and runner shorthand;
  - dependency graph edges;
  - duplicate ids;
  - unknown/self dependencies;
  - cycles;
  - issue accumulation;
  - trigger filter parsing/type-checking deferral.

`WorkflowModel` deliberately does not contain project ids, commit shas, refs, dates, DB rows, source paths, or other platform metadata. Those belong to `WorkflowDefinition`.

Trigger filters remain source strings in this PR. They will become typed workflow expressions only after event-schema-backed expression contexts are introduced.

## Validation

Passed:

- `mise exec -- turbo build --filter=@shipfox/api-definitions`
- `mise exec -- turbo type --filter=@shipfox/api-definitions`
- `mise exec -- turbo check --filter=@shipfox/api-definitions`
- `node .claude/skills/readme-writer/scripts/readability.mjs libs/api/definitions/src/core/workflow-model/README.md`
- Pure DB-free Vitest config for workflow YAML/model tests: 2 files, 23 tests passed

Blocked:

The branch-level test run reaches API package global setup and fails because Postgres is unavailable. Docker is not running in this environment, and `docker compose up -d` fails because the Docker daemon socket is missing.

## Review

Claude consult review was run for this rewritten PR. Accepted recommendations were applied:

- added issue accumulation coverage;
- added stable-id edge coverage;
- added top-level runner string shorthand coverage;
- added single manual trigger coverage;
- standardized tmp architecture notes on the `WorkflowModel*` naming style;
- deferred the one-member `acceptance` model field until gates add real acceptance semantics;
- recorded the future trigger-filter shape change as deferred work outside the repo.

## Stack

Part 1 of 4.

Depends on: none
Next: #129

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #128 | `workflow-model-ir` | `main` |
| 2 | #129 | `workflow-definitions-integration` | `workflow-model-ir` |
| 3 | #130 | `workflow-runtime-scheduling` | `workflow-definitions-integration` |
| 4 | #131 | `workflow-runtime-kernel-spike` | `workflow-runtime-scheduling` |


noe-charmet:
I don't think we specifically need a shared library for internal representations. Per our DDD principles, each module should own its own representation. How the `trigger` module sees a workflow and how other modules see it is each module's own business.

We already have two modules that own a representation of this: `definition` and `workflow`.

- The `definition` core entities are literally how we represent a workflow definition (i.e., transcribing the YAML into code) and store it in the DB. It looks like this mostly overlaps with what's being proposed here. Maybe you just need to update those core entities to match the new representation?
- The `workflow` core entities represent an actual workflow run (i.e., one execution of a workflow). They share a lot with the definitions but take on more coordination responsibilities. This is typically the step where we'd run expression interpolation and evaluation.

So it seems like we don't need or want a shared "internal representation." Instead, we have modules that each own their own internal representation as their "core" entities. Each module also owns its persistence and presentation layers, and no other module should have business logic that reaches into another module's entities.

Here's how I'd picture the data path and how things would be structured:

1. The `definition` module receives the source YAML string. It parses it and stores its own representation of the definition. No interpolation or expression evaluation happens at this step.
2. The `trigger` module sees a new definition and updates its representation of the available triggers and their respective definition IDs. Still no expression evaluation or interpolation here.
3. The `trigger` module receives an integration event and tries to match it against the triggers in its storage (based on project ID / source / event). If the matched trigger has a conditional expression, it evaluates that expression to confirm whether it should run. This is the first time we actually run an evaluation, and it uses the context available to the `trigger` module.
4. For a confirmed match, the `trigger` module calls the `workflow` module to launch a workflow, passing along the event context and the definition ID.
5. The `workflow` module fetches the definition from the `definition` module and builds its own representation of the workflow run. As it runs and decides which job comes next, it evaluates job conditions when needed. When an `if` is present, that evaluation uses the full context of the workflow run.

The point I'm hoping to highlight is that each module keeps its own representation, and that we probably don't want an external shared library for this (and if we ever do decide we want one, we should scope it to the api libs). You'll also notice that I think we should keep our evaluatable expressions in their source representation until the very last moment, only resolving them when we actually need to evaluate.

noe-charmet (dismissed):

PR #129, opened 2026-06-08
Integrate definitions with workflow model validation

ThierryAbalea:
## Stack

Part 1 of 3.

Depends on: none
Next: #130

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #129 | `workflow-definitions-integration` | `main` |
| 2 | #130 | `workflow-runtime-scheduling` | `workflow-definitions-integration` |
| 3 | #131 | `workflow-runtime-kernel-spike` | `workflow-runtime-scheduling` |

## Objective

Integrate `libs/api/definitions` with the accepted workflow document and normalized `WorkflowModel` pipeline.

Definitions now parse YAML through `core/workflow-yaml`, normalize it through `core/workflow-model`, and persist/expose a `WorkflowDefinition` shape that contains both:

- `document`: the accepted authoring document, useful for auditability and dashboard display;
- `model`: the normalized semantic model used by downstream definitions/runtime integration.

## Notable Changes

- Replaces the old internal `WorkflowSpec` entity with `WorkflowDefinitionPayload`.
- Keeps the existing `workflow_definitions.definition` JSONB column name, but stores `{document, model}` in that column.
- Removes the old `validateDag` helper; dependency validation and cycle detection now live in `normalizeWorkflowDocument`.
- Updates create/validate definition routes to return/use `definition` rather than `spec`.
- Renames the route error code from `invalid-definition-spec` to `invalid-workflow-definition`.
- Removes the dead trigger `on` field from definitions DTO trigger schema.
- Adds pure tests for outbox trigger projection and `toDefinitionDto`.

## Files And Modules

- `libs/api/definitions/src/core/entities/workflow-definition.ts`
- `libs/api/definitions/src/core/parse-definition.ts`
- `libs/api/definitions/src/core/validate-definition.ts`
- `libs/api/definitions/src/db/schema/definitions.ts`
- `libs/api/definitions/src/db/definitions.ts`
- `libs/api/definitions/src/db/definition-triggers.ts`
- `libs/api/definitions/src/presentation/dto/definition.ts`
- `libs/api/definitions/src/presentation/routes/create-definition.ts`
- `libs/api/definitions/src/presentation/routes/validate-definition.ts`
- `libs/api/definitions/src/core/sync-definitions.ts`
- `libs/api/definitions/src/temporal/activities/sync-activities.ts`
- `libs/api/definitions-dto/src/schemas/dto.ts`
- `libs/api/definitions-dto/src/schemas/trigger.ts`

## Tests

Passed:

```sh
mise exec -- turbo type --filter=@shipfox/api-definitions --filter=@shipfox/api-definitions-dto
mise exec -- turbo check --filter=@shipfox/api-definitions --filter=@shipfox/api-definitions-dto
mise exec -- ./tools/vitest/node_modules/.bin/vitest run --config libs/api/definitions/vitest.pure.config.ts
mise exec -- turbo build --filter '...[origin/main]'
mise exec -- turbo check --filter '...[origin/main]'
```

The temporary pure Vitest config covered 7 files / 45 tests and was deleted after the run.

Blocked by local environment:

```sh
mise exec -- turbo test --filter=@shipfox/api-definitions
mise exec -- turbo test --filter '...[origin/main]'
```

Both fail in `libs/api/definitions/test/globalSetup.ts` because PostgreSQL is not reachable on `::1:5432` or `127.0.0.1:5432`.

## Notes

This PR intentionally does not add a DB migration because the JSONB column name and SQL type stay unchanged. It assumes this workflow-definition data shape is not yet deployed with durable production rows. If an environment already has old `WorkflowSpec` JSON in `workflow_definitions.definition`, it must be invalidated, resynced, or backfilled before deploying this stack.

Trigger filter CEL parsing remains deferred. `WorkflowModelTrigger.filter` is still a source string until event-schema-backed CEL validation exists.


libs/api/definitions-dto/src/schemas/dto.ts
```diff
@@ -29,7 +29,10 @@ export const definitionDtoSchema = z.object({
   sha: z.string().nullable(),
   ref: z.string().nullable(),
   name: z.string(),
-  definition: z.record(z.string(), z.unknown()),
+  definition: z.object({
+    document: z.unknown(),
+    model: z.unknown(),
+  }),
```
noe-charmet: Do we want to expose both to users ?
IMO we should decide on one representation that is shared outside of this module. I'd tend to think that the document is what end-users want to see.

Even if we were to share both, I think we don't need to nest them, but we should have them as top-level properties. The `definition` property of a `WorkflowDefinition` being an object with sub-properties is a bit weird in terms of semantics. Especially as they are top-level fields on the core entity.

libs/api/definitions/src/core/entities/workflow-model.ts
```diff
@@ -12,6 +12,7 @@ export interface WorkflowModelTrigger {
   readonly source: string;
   readonly event: string;
   readonly inputs?: Readonly<Record<string, unknown>>;
+  // Trigger filters stay as source strings until event-schema-backed CEL validation exists.
```
noe-charmet: 🤖 💬

ThierryAbalea: Done. I removed the redundant inline trigger-filter note and kept the behavior documented in the workflow-model README.

libs/api/definitions-dto/src/schemas/dto.ts
ThierryAbalea: Done. I flattened the public DTO shape: definition responses now expose `workflow_document` and `workflow_model` as top-level fields instead of `definition: { document, model }`. The DB/core payload remains unchanged internally.

PR #141, opened 2026-06-10
Expand AGENTS.md error-handling guidance for agents

noe-charmet:
Expand and restructure the error-handling guidance in AGENTS.md so agents
follow the repo's existing pattern instead of inventing one.

The previous guidance was a single three-sentence paragraph nested under HTTP
routes: it stated the principle (throw typed domain errors, translate to
`ClientError` at the route edge) but gave nothing concrete to pattern-match
against, and its placement made error handling read like a presentation-only
detail.

Error handling actually spans every backend layer — core, db, api,
presentation, and the global handler — so this promotes it to its own top-level
section organized by layer:

- **Domain errors (`core` / `db`)** — typed `Error` subclasses in
  `core/errors.ts` with no HTTP concerns, reusable by workers/subscribers/jobs.
- **External and infrastructure errors (`api` / `core`)** — catch external SDK
  errors and re-throw as typed errors carrying a `reason` enum, preserving
  `cause`.
- **Translating to HTTP responses (`presentation`)** — per-route `errorHandler`
  to `ClientError`, the `ClientError` params (`status` defaulting to 400,
  client-safe `details` vs log-only `data`, `cause`), and `reason` → status
  mapping.
- **The global handler** — what it sends and what it reports to Sentry.

Two notes for review:
- This documents the local repo's pattern, which diverges from the reference
  doc in another Shipfox checkout (`defineRoute`/Zod here vs
  `FastifyPluginCallback`/Typebox there) — so it's not a copy.
- The `cause` guidance is slightly ahead of current practice: most
  domain→`ClientError` conversions currently drop the cause. Kept it because it
  improves Sentry context; happy to soften it to match existing reality if
  preferred.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -210,9 +210,10 @@ throw new IntegrationProviderError('rate-limited', message, retryAfterSeconds);
 
 ### Translating to HTTP responses (`presentation`)
 
-Routes are the only place that turns errors into client responses. Translate
-known errors to `ClientError` in the route's `errorHandler`, and re-throw
-anything you don't recognize so it reaches the global handler:
+The presentation/request boundary is the only place that turns errors into
+client responses. Route `errorHandler`s are the default place to translate
+known domain errors to `ClientError`; re-throw anything you don't recognize so
+it reaches the global handler:
 
 ```ts
 errorHandler: (error) => {
@@ -232,7 +233,8 @@ structured client response. `code` is a stable kebab-case string (`not-found`,
 - `details` — structured data **returned to the client** (snake_case keys).
 - `data` — context **logged only**, never sent to the client.
 - `cause` — pass the original error when wrapping, so it stays in the chain for
-  Sentry.
+  logs and diagnostics. Sentry sees unknown errors that reach the global handler;
+  handled cases must capture explicitly if they need Sentry reporting.
 
 For external errors, map the `reason` enum to a status and client code here
 (e.g. `rate-limited` → 429).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,17 +60,70 @@ src/
   presentation/  Fastify routes, auth adapters, and DTO conversion
 ```
 
-### HTTP routes and errors
-
-Domain and persistence code should throw typed domain errors. Translate those
-errors to `ClientError` only at the Fastify route edge with stable client-facing
-error codes and HTTP statuses. Unexpected errors should be allowed to reach the
-shared error handler.
+### HTTP routes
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -169,6 +164,87 @@ Each E2E package must also declare an explicit workspace dependency on the packa
...
+```
+
+`ClientError(message, code, params)` is the only error that produces a
+structured client response. `code` is a stable kebab-case string (`not-found`,
+`last-member`, `project-already-exists`). `params`:
+
+- `status` — HTTP status, **defaults to 400**; set it for any other 4xx.
+- `details` — structured data **returned to the client** (snake_case keys).
+- `data` — context **logged only**, never sent to the client.
+- `cause` — pass the original error when wrapping, so it stays in the chain for
```
ThierryAbalea, line 235: `cause` is worth keeping in the guidance, but “for Sentry” seems too strong today: handled `ClientError`s are logged and returned by the global handler, while `captureException(...)` only runs for unknown 500s. Could we phrase this as preserving the original error for logs/diagnostics, or explicitly note that Sentry only sees unknown errors unless a handled case captures separately?

AGENTS.md
```diff
@@ -169,6 +164,87 @@ Each E2E package must also declare an explicit workspace dependency on the packa
...
+Callers then branch on `reason` without depending on the SDK's error shapes.
+Pass `retryAfterSeconds` for backpressure reasons like `rate-limited`:
+
+```ts
+throw new IntegrationProviderError('rate-limited', message, retryAfterSeconds);
+```
+
+### Translating to HTTP responses (`presentation`)
+
+Routes are the only place that turns errors into client responses. Translate
```
ThierryAbalea: “Routes are the only place” feels a bit too narrow compared to the current codebase: auth methods and request helpers like `createJwtAuthMethod()` / `requireMembership()` also throw `ClientError` directly, which seems fine because they live at the HTTP request boundary. Could we phrase this as “the presentation/request boundary is the only place...” and keep route `errorHandler`s as the default place for translating domain errors?

ThierryAbalea (dismissed):

AGENTS.md
noe-charmet: Done in 3e839da. I changed this to describe the presentation/request boundary as the HTTP translation edge, and kept route `errorHandler`s as the default place for translating known domain errors.

AGENTS.md
noe-charmet: Done in 3e839da. I rephrased `cause` as preserving the original error for logs/diagnostics, and clarified that Sentry only sees unknown errors that reach the global handler unless a handled case captures explicitly.

PR #143, opened 2026-06-10
Make runner scheduling step-less

noe-charmet:
Make scheduling step-less so runner claims bind only a job/run lease instead of transporting step payloads through the runner queue.

This moves pending runner work to `{job_id, run_id, workspace_id}`, exposes `scheduleJob(...)`, and makes the claim route mint and return `{job_id, run_id, lease_token}`. Stuck running jobs now expire the lease and emit `runners.job.lease_expired`, leaving failure handling to ENG-400 rather than fabricating an empty-steps completion.

Review notes: `libs/api/runners/drizzle/0000_initial.sql` and its snapshot were edited in place because no runner migration has shipped. Existing local DBs need to be recreated because the migration hash changes, for example with `docker compose down -v && docker compose up -d` or an equivalent targeted drop.

Until ENG-399, existing runner agents compile but fail runtime parsing of the new claim response. Until ENG-400 wires the lease-expired subscriber, `runners.job.lease_expired` is intentionally zero-subscriber plumbing and stuck jobs are only failed by the 60-minute workflow backstop. There is also still an at-least-once enqueue window where a retry after claim can reinsert a pending row; ENG-400 will document or close that contract.

Fixes ENG-398

---
## Summary by cubic
Make runner scheduling step-less and token-based: we now queue only ids; runners claim a lease token, and stuck leases emit `runners.job.lease_expired`. Implements ENG-398.

- **New Features**
  - Replaced `enqueueJob` with `scheduleJob({jobId, runId, workspaceId})` in `@shipfox/api-runners` (idempotent on `jobId` while pending).
  - `/runners/jobs/request` now returns `{job_id, run_id, lease_token}` (`claimedJobResponseSchema` in `@shipfox/api-runners-dto`).
  - Added `RUNNER_JOB_LEASE_EXPIRED` (`runners.job.lease_expired`) event; the detector now expires leases and emits this event.
  - Removed `payload` from `runners_pending_jobs`; pending work stores only ids.
  - Workflows: removed step payloads from scheduling (`enqueueJobForRunner`, `JobOrchestrationInput`) to keep scheduling step-less end to end.

- **Migration**
  - Recreate local runner DBs (initial migration edited), e.g. `docker compose down -v && docker compose up -d`.
  - Replace `enqueueJob` calls with `scheduleJob`; update agents to parse the new claim response (no steps; use `lease_token`).
  - Until ENG-399/ENG-400: older agents will fail parsing; `runners.job.lease_expired` has no subscriber and the 60‑min backstop still handles failures.

<sup>Written for commit 93f97319cf8f6459d2a84d2fe2e88a8cead0efda. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libs/api/runners/src/db/jobs.ts
+++ b/libs/api/runners/src/db/jobs.ts
@@ -18,9 +18,12 @@ export interface ScheduleJobParams {
   runId: string;
 }
 
-// Idempotent: a duplicate jobId means the job is already scheduled. Temporal
-// retries the enqueue activity at-least-once, so a unique-violation throw on a
-// retry-after-lost-result would permanently fail a healthy job's workflow.
+// Idempotent while the job is still pending: a duplicate jobId already in
+// `runners_pending_jobs` is a no-op. Temporal retries the enqueue activity
+// at-least-once, so a unique-violation throw on a retry-after-lost-result
+// would permanently fail a healthy job's workflow. This does not cover a retry
+// after the job has been claimed and moved to `runners_running_jobs`; that
+// window is owned by ENG-400.
 export async function scheduleJob(params: ScheduleJobParams): Promise<void> {
   await db()
     .insert(pendingJobs)
--- a/libs/api/runners/src/db/jobs.ts
+++ b/libs/api/runners/src/db/jobs.ts
@@ -21,9 +21,10 @@ export interface ScheduleJobParams {
 // Idempotent while the job is still pending: a duplicate jobId already in
 // `runners_pending_jobs` is a no-op. Temporal retries the enqueue activity
 // at-least-once, so a unique-violation throw on a retry-after-lost-result
-// would permanently fail a healthy job's workflow. This does not cover a retry
-// after the job has been claimed and moved to `runners_running_jobs`; that
-// window is owned by ENG-400.
+// would permanently fail a healthy job's workflow. The guard does not extend
+// past the claim: once the job has moved to `runners_running_jobs`, a retry
+// can reinsert a pending row that a later claim then rejects on the
+// running-job unique constraint.
 export async function scheduleJob(params: ScheduleJobParams): Promise<void> {
   await db()
     .insert(pendingJobs)
@@ -126,8 +127,9 @@ export async function finalizeRunningJob(
  * written in the same transaction only when a row was deleted, so a heartbeat
  * landing between an outer `findStuckJobs` read and this call leaves the live
  * job untouched, and overlapping detector runs cannot double-emit. `runId`
- * comes from `RETURNING`. (Duplicates `finalizeRunningJob`'s shape on purpose;
- * `finalizeRunningJob` is removed in ENG-400.)
+ * comes from `RETURNING`. Intentionally not shared with `finalizeRunningJob`:
+ * the two guard on different predicates and emit different events, so the
+ * resemblance is shallow.
  */
 export async function expireStuckJob(params: {
   jobId: string;
--- a/libs/api/runners/src/db/jobs.ts
+++ b/libs/api/runners/src/db/jobs.ts
@@ -1,6 +1,6 @@
 import {
-  type JobPayloadDto,
   RUNNER_JOB_COMPLETED,
+  RUNNER_JOB_LEASE_EXPIRED,
   type RunnersEventMap,
   type StepResultDto,
 } from '@shipfox/api-runners-dto';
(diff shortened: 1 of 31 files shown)


libs/api/runners/src/db/jobs.ts
```diff
@@ -12,38 +12,43 @@ import {runnersOutbox} from './schema/outbox.js';
...
 }
 
-export async function enqueueJob(params: EnqueueJobParams): Promise<void> {
-  await db().insert(pendingJobs).values({
-    workspaceId: params.workspaceId,
-    jobId: params.jobId,
-    runId: params.runId,
-    payload: params.payload,
-  });
+// Idempotent: a duplicate jobId means the job is already scheduled. Temporal
```
ThierryAbalea: This reads a bit broader than the implementation: `scheduleJob` is idempotent while the duplicate `job_id` is still in `runners_pending_jobs`, but not after the job has been claimed and moved to `runners_running_jobs`. Could we narrow the comment to make that scope explicit?

libs/api/runners/src/db/jobs.ts
```diff
@@ -12,38 +12,43 @@ import {runnersOutbox} from './schema/outbox.js';
...
+// retry-after-lost-result would permanently fail a healthy job's workflow.
+export async function scheduleJob(params: ScheduleJobParams): Promise<void> {
+  await db()
+    .insert(pendingJobs)
+    .values({
+      workspaceId: params.workspaceId,
+      jobId: params.jobId,
+      runId: params.runId,
+    })
+    .onConflictDoNothing({target: pendingJobs.jobId});
```
ThierryAbalea, line 36: Do you already plan to close the retry-after-claim window in ENG-400? A retry after claim can recreate a pending row for a job that is already running, and the next claim will then hit the `running_jobs.job_id` unique constraint. Just want to confirm whether that is part of the planned follow-up or should be handled here.

libs/api/runners/src/presentation/routes/request-job.ts
```diff
@@ -33,8 +33,7 @@ export const requestJobRoute = defineRoute({
     return {
       job_id: job.jobId,
       run_id: job.runId,
-      job_name: job.payload.job_name,
-      steps: job.payload.steps,
+      lease_token: job.leaseToken,
```
ThierryAbalea, line 36: The PR body already calls out ENG-399, so this is fine if the stack lands together. If this PR can merge before ENG-399, though, the current runner app still parses `/runners/jobs/request` with `jobPayloadResponseSchema`, which requires `job_name` and non-empty `steps`; this new `{job_id, run_id, lease_token}` response will fail runtime parsing in `apps/runner/src/api-client.ts`. Should this stay draft-blocked until the runner consumer is updated?

ThierryAbalea (dismissed):

libs/api/runners/src/db/jobs.ts
noe-charmet: Good call — narrowed the comment to make the scope explicit: `scheduleJob` is idempotent only while the job is still in `runners_pending_jobs`, and the comment now says so and notes that the retry-after-claim window (job already moved to `runners_running_jobs`) is owned by ENG-400.

libs/api/workflows/src/temporal/activities/orchestration-activities.ts
```diff
@@ -144,18 +144,10 @@ export async function enqueueJobForRunner(params: {
...
 }): Promise<void> {
-  const payload: JobPayloadDto = {
-    job_id: params.jobId,
-    run_id: params.runId,
-    job_name: params.jobName,
-    steps: params.steps,
-  };
-
-  await enqueueJob({
+  await scheduleJob({
```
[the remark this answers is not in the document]
noe-charmet: Fixed. Removed the dead `jobName` and `steps` params from `enqueueJobForRunner` so the scheduling activity only carries `{workspaceId, jobId, runId}`. Since those fields were no longer consumed anywhere downstream, I also dropped them from the rest of the chain — `JobOrchestrationInput` and the `launchJobs` call site in `run-orchestration.ts` — so no stale step-content transport remains. The test harness now sources the simulated runner step report from the DAG (`loadRunDag`) instead of the enqueue params, matching the real step-less flow.

libs/api/runners/src/db/jobs.ts
noe-charmet: Confirmed — this is planned follow-up, not handled here. `scheduleJob` is intentionally only idempotent while the row is still pending; the retry-after-claim window (a retry recreating a pending row for a job already in `runners_running_jobs`, which then collides on the `running_jobs.job_id` unique constraint) is left as a known at-least-once contract for ENG-400 to either close or document. I've narrowed the comment on `scheduleJob` to make that scope explicit (pushed in f5a95b3).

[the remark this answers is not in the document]
noe-charmet: Valid observation, and intentional for this PR. `scheduleJob` is idempotent only while the job is still pending — once it has moved to `runners_running_jobs`, a retry can recreate a pending row and a later claim will conflict on `running_jobs.job_id`. This at-least-once enqueue window is a known contract that ENG-400 will close or document; it is called out in the PR description. I've narrowed the `scheduleJob` comment to state this scope explicitly (f5a95b3) so it is not mistaken for full by-jobId idempotency.

libs/api/runners/src/presentation/routes/request-job.ts
noe-charmet: Right — this is meant to land as a stack with ENG-399, not standalone. The new `{job_id, run_id, lease_token}` claim response would fail runtime parsing in `apps/runner/src/api-client.ts` (still expecting `job_name` + non-empty `steps`) if this merged first, so it stays blocked until ENG-399 updates the runner consumer. Keeping the stack ordered avoids a window where deployed runners can't parse the claim response.

libs/api/runners/src/db/jobs.ts
```diff
@@ -12,38 +12,47 @@ import {runnersOutbox} from './schema/outbox.js';
...
 }
 
-export async function enqueueJob(params: EnqueueJobParams): Promise<void> {
-  await db().insert(pendingJobs).values({
-    workspaceId: params.workspaceId,
-    jobId: params.jobId,
-    runId: params.runId,
-    payload: params.payload,
-  });
+// Idempotent while the job is still pending: a duplicate jobId already in
```
ThierryAbalea, line 21: Could we close this retry-after-claim window here rather than leave it only to ENG-400? The current `onConflictDoNothing` is safe while the row is still pending, but if Temporal retries `scheduleJob` after the job was claimed, it can create a new pending row for a job already in `runners_running_jobs`. Then `claimPendingJob` selects that FIFO row, the running-job insert hits the `job_id` unique constraint, the transaction rolls back the delete, and the pending duplicate stays as a poison row. Once the original running row is finalized, that duplicate can be claimed and re-run the completed job. Could `claimPendingJob` absorb this case, for example with a conflict-aware running insert that deletes the duplicate pending row and returns `null`, plus a `schedule -> claim -> schedule same jobId` regression test?

ThierryAbalea (approved):
Approved with one follow-up comment on the retry-after-claim idempotency window.

PR #145, opened 2026-06-10
[api/workflows] Add runner-facing step endpoints with lease-token auth

noe-charmet:
Adds the agent-facing per-step HTTP surface owned by Orchestration: the runner agent pulls its next step and reports results against the job named by its lease token, with the token verified in-process so the hot path touches only Postgres plus a local signature check.

This is the third leg of the per-step runner protocol (after the lease-token codec and the step-progression domain service): without it the agent has no HTTP surface to drive a job step by step.

- `POST /runs/jobs/current/steps/next` → `200 {kind:'step', step}` (re-delivers the in-flight step on retries) or `200 {kind:'done', status}` once the job is terminal; `404` is reserved for unknown jobs so the agent can distinguish "job over" from "bad id".
- `POST /runs/jobs/current/steps/:stepId/report` → records a result idempotently and returns `{ok, cancel}`, where `cancel` tells the agent to stop because the job finished without full success; `404` unknown/cross-job step, `409` never-dispatched step.
- The `lease-token` auth method derives the job from the token claims, so the URLs carry no job id — the planned `:jobId` param plus 403 mismatch guard were dropped as redundant with the capability token.

Two preparatory refactors ride along in their own commits:

- The job lease token codec (issue/verify, claims schema, secret config) moves from `api-runners`/`api-runners-dto` into `api-auth`/`api-auth-dto` — token signing/verification is an authentication concern, and Scheduling (ENG-398) will mint through `api-auth` the same way Orchestration verifies. Env vars renamed `RUNNERS_JOB_LEASE_TOKEN_*` → `AUTH_JOB_LEASE_TOKEN_*`.
- `extractBearerToken` was about to gain a third copy-paste, so it moves into `@shipfox/node-fastify` next to the auth method registry; both existing auth adapters now share it.

Worth careful review: the `cancel` derivation (`jobFinished && status === 'failed'`) and the trust boundary in `lease-token-auth.ts` — `runId`/`workspaceId` claims are intentionally informational and not verified against the DB (to be documented in ENG-401). Route tests cover the auth-negative matrix, idempotency/never-downgrade, snake_case→camelCase error round-trip, and route-level concurrency against real Postgres.

Refs ENG-397

---
## Summary by cubic
Adds runner-facing per-step endpoints secured by a job lease token so agents can pull the next step and report results safely. Moves the lease-token codec and auth method into `@shipfox/api-auth` (registered as `leased-job`), shares a leased‑job request context, adds `extractBearerToken` in `@shipfox/node-fastify`, clarifies OpenAPI (step endpoints do not finalize jobs), tightens validation, and fixes the step error signal round‑trip (ENG-397, ENG-401).

- New Features
  - POST `/runs/jobs/current/steps/next` → returns the next pending step, or `{kind:'done', status}` when the job is finished; 404 for unknown jobs.
  - POST `/runs/jobs/current/steps/:stepId/report` → idempotently records a step result and returns `{ok, cancel}`; 404 for unknown/cross-job step, 409 for never-dispatched; validates payloads (failed requires error, succeeded forbids error).
  - Lease-token auth: locally verifies a Bearer token, sets a `leased-job` request context, and derives the job from claims (no `:jobId` in URLs); provided and auto-registered by `@shipfox/api-auth`.
  - Docs: per-step routes do not finalize the job; runners must still call the job-completion endpoint.

- Migration
  - Rename env vars: `RUNNERS_JOB_LEASE_TOKEN_*` → `AUTH_JOB_LEASE_TOKEN_*`.
  - Lease-token codec moved to `@shipfox/api-auth`/`@shipfox/api-auth-dto`; update imports to `issueJobLeaseToken`/`verifyJobLeaseToken` from `@shipfox/api-auth` and claims/types from `@shipfox/api-auth-dto`.
  - If you had a local lease-token auth adapter, remove it and use the registered `leased-job` method (`auth: AUTH_LEASED_JOB` on routes).

<sup>Written for commit 968f4cc2797a75fa02dbaa5883f181ad7be825ed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/145?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libs/api/workflows/src/presentation/routes/report-step.ts
```diff
@@ -0,0 +1,43 @@
...
+    const leasedJob = requireLeasedJobContext(request);
+
+    const outcome = await recordStepResult({
+      jobId: leasedJob.jobId,
+      stepId,
+      status: request.body.status,
+      error: fromStepErrorDto(request.body.error),
+    });
+
+    return {ok: true, cancel: outcome.jobFinished && outcome.status === 'failed'};
```
ThierryAbalea, line 41: The new per-step endpoints can drive all step rows to terminal, but they do not appear to finalize the job or signal the Temporal `job-completed` workflow. The existing workflow still only marks the job final after receiving that signal, and the current runner still sends it through `/runners/jobs/:jobId/complete`.

If the intended protocol is 'pull/report each step, then still call the legacy complete endpoint,' could we document that explicitly in the route/PR description? Otherwise a runner that only follows the new step API could finish every step, see `{kind: 'done'}`, and still leave the job `running` until the timeout path fails it.

libs/api/auth/src/presentation/auth/lease-token-auth.ts
```diff
@@ -0,0 +1,29 @@
+import {AUTH_LEASED_JOB, setLeasedJobContext} from '@shipfox/api-auth-context';
+import {type AuthMethod, ClientError, extractBearerToken} from '@shipfox/node-fastify';
+import {verifyJobLeaseToken} from '#core/job-lease-token.js';
+
+/**
```
ThierryAbalea, line 5: This auth path intentionally treats the signed lease token as the authority, so `runnerTokenId` is not checked against the DB for revocation. That seems defensible for a job-scoped 90-minute capability token, but it is an important security tradeoff: a lease minted before runner-token revocation remains usable until expiry.

Could we document that revocation window near the trust-boundary comment, or note the planned DB check/status binding if this is expected to tighten later?

ThierryAbalea (dismissed):
Approved with two non-blocking comments on the step-protocol finalization contract and lease-token revocation tradeoff.

libs/api/workflows/src/presentation/routes/report-step.ts
noe-charmet: Good catch — you're right, and the intended protocol is exactly "pull/report each step, then still call the legacy complete endpoint." Finalization deliberately stays on the runner job-completion path (`POST /runners/jobs/:jobId/complete` → Temporal `job-completed` signal); the step endpoints only advance and record steps. The per-step path takes over finalization at the ENG-400 cutover.

Documented explicitly in the route descriptions (968f4cc):
- `report-step`: "Reporting every step does not by itself finalize the job — the runner must still call the job-completion endpoint, which is what drives the job to its terminal state."
- `next-step`: the "no more steps" response now states it "does not finalize the job, which the runner still completes through the job-completion endpoint" (instead of implying the job has finished).

So a runner that only follows the step API and stops at the done response is misusing the protocol; the docs now make the required completion call unambiguous.

libs/api/auth/src/presentation/auth/lease-token-auth.ts
noe-charmet: Agreed — it's a deliberate tradeoff and now documented in two places (642f7a1):

- The trust-boundary comment here in `lease-token-auth.ts` is expanded to spell out the revocation window: `runnerTokenId` is carried but not checked against the runner-token table, so a lease minted just before its runner token is revoked stays usable until it expires (`AUTH_JOB_LEASE_TOKEN_EXPIRES_IN`, 90m). It states why it's accepted (short-lived, single-job capability; a per-request DB lookup would sit on the hot status-reporting path) and the tightening path (bind to runner status).
- A new "Auth methods" section in the `api-auth` README covers both auth methods side by side, including a "Tradeoff — no revocation" writeup for the lease token with the same rationale, so the full reasoning lives next to the user-JWT stale-membership tradeoff.

No DB check is planned for now; the comment notes to add a status binding only if the window proves too wide.

PR #146, opened 2026-06-10
[api/integration/github] Test each package's webhook concerns in its owner

noe-charmet:
Restructures the GitHub integration test tooling to follow the Fishery-factory + db-layer conventions used across the other API packages, and to keep github's and core's test concerns within their owning packages.

## Problem

github's tooling reached the core-owned `integrations_*` tables through raw SQL in a single `test/core-fixtures.ts` file. That diverged from how the other API packages seed/read test data, and — more importantly — made github's tests depend on **core's** schema. The github webhook route doesn't own connection/outbox/delivery persistence; it calls injected functions that `@shipfox/api-integration-core` implements and wires in production. Testing that persistence inside github forced github to redeclare core's tables.

## Change

Split the tests along ownership:

- **github** route tests fake the injected core functions (`publishIntegrationEventReceived`, `recordDeliveryOnly`, `getIntegrationConnectionById`) with `vi.fn()` spies — the same interface seam `install.test.ts` already fakes. They assert what the route owns: signature/payload validation and which core function it calls with what mapped event. They touch only github's own `installations` table, seeded via `githubInstallationFactory`.
- **core** gains a persistence test exercising `publishIntegrationEventReceived` / `recordDeliveryOnly` directly against its own tables (row shape, dedup, idempotent delivery recording).

Each package's tests now touch only its own schema — no redeclared tables, no raw SQL, no cross-package test coupling.

Supporting changes:
- **Fishery factory** (`githubInstallationFactory`) persists through github's `#db` layer.
- **Raw webhook payloads** move to `test/fixtures/`.
- **Drop the `integrations_github_installations -> integrations_connections` foreign key.** The cascade was unused, and the FK was the only thing forcing github's tests to know core's `connections` table. Edited the existing migration since nothing is deployed.

## Reviewer notes
- No runtime behavior change.
- The full HTTP-to-rows path (route → real core persistence) is intentionally not re-tested in one place; it's covered piecewise (github asserts the call + mapping, core asserts the persistence) and belongs to app-level E2E.
- Test-tooling only, so no changeset.


libs/api/integration/github/src/presentation/routes/webhooks.test.ts
```diff
@@ -1,52 +1,61 @@
...
+  app: FastifyInstance;
+  publishIntegrationEventReceived: ReturnType<typeof vi.fn>;
+  recordDeliveryOnly: ReturnType<typeof vi.fn>;
+  getIntegrationConnectionById: ReturnType<typeof vi.fn>;
+}
+
+async function createTestApp(options: {connection?: IntegrationConnection} = {}): Promise<TestApp> {
+  const publishIntegrationEventReceived = vi.fn(() => Promise.resolve({published: true}));
+  const recordDeliveryOnly = vi.fn(() => Promise.resolve());
+  const getIntegrationConnectionById = vi.fn(() =>
```
ThierryAbalea, line 43: The valid-push test seeds an installation with the factory default random `connectionId`, then creates a separate `fakeConnection()` with another random `id`. Since `getIntegrationConnectionById` ignores its id argument and always returns the configured connection, this test no longer proves the route looks up the connection associated with the installation row. Could we seed the installation with `connectionId: connection.id` and assert `getIntegrationConnectionById` was called with that id and a transaction? Or make the fake return the connection only when the requested id matches. That would preserve the package ownership split while keeping this seam constrained.

ThierryAbalea, line 43: Related: after moving persistence behind spies, do we still cover the branch where the installation exists but `getIntegrationConnectionById` returns `undefined`? That path should record the delivery only and skip publishing. Since dropping the FK makes dangling installation rows possible by schema, a small route/core test for that case would make the intended behavior explicit.

ThierryAbalea (dismissed):
Approved. The ownership split makes sense; I left two non-blocking test-coverage comments around constraining the connection lookup fake and covering the dangling-installation path after dropping the FK.

libs/api/integration/github/src/presentation/routes/webhooks.test.ts
noe-charmet: Done in 30df33e. The valid-push test now seeds the installation with `connectionId: connection.id` and asserts `getIntegrationConnectionById` is called once with that id and a tx, so it proves the route looks up the connection from the installation row.

noe-charmet: Done in 30df33e. Added a route test 'records the delivery only when the installation has no connection' that makes the connection lookup return undefined for an existing installation and asserts the delivery is recorded once while publishing is skipped.

PR #147, opened 2026-06-10
[api/integrations] Add Sentry webhook ingest provider

noe-charmet:
Adds a Sentry integration provider that ingests Sentry **issue** webhooks and turns them into workflow triggers. This is **PR1 of 2** — webhook ingest only; the install/connect flow follows in PR2.

## Why

We want workflows to be triggerable by Sentry issue events. The triggers/workflows seam is already fully source-agnostic (no `(source, event)` whitelist, generic `TriggerPayload` variant, passing sentry fan-out test), so this PR only adds a provider that receives webhooks, normalizes the payload, and publishes `integrations.event.received` with `source: 'sentry'` and `event: issue.<action>`. The triggers subscriber fans each event out to every matching subscription in the workspace.

## What

- **`@shipfox/api-integration-sentry` + `-sentry-dto` (new):** webhook receiver with manual HMAC-SHA256 verification (length-guarded `timingSafeEqual`), `Request-ID` dedup against the `(provider, deliveryId)` PK, issue normalization to a stable `SentryIssuePayload`, and an `installation` lifecycle handler (uninstall disables the connection). A raw `ignored` action is normalized to `archived`.
- **`@shipfox/api-integration-core-dto`:** `SentryIssuePayload` contract.
- **`@shipfox/api-integration-core`:** register Sentry behind `INTEGRATIONS_ENABLE_SENTRY_PROVIDER`; add `updateIntegrationConnectionLifecycleStatus`; pin stable per-provider migration-tracking table names.
- **`@shipfox/node-fastify`:** shared `rawBodyPlugin` + `WEBHOOK_BODY_LIMIT` exports (placed here, not integration-core, to avoid a `core ↔ provider` dependency cycle).
- **`@shipfox/node-module`:** optional `ModuleDatabase.migrationsTableName` so conditionally-composed provider databases get a position-independent migration table.
- **`@shipfox/api-integration-github`:** consume the shared `rawBodyPlugin` instead of a local copy (internal refactor, no behavior change).

## Deliberate decisions worth a careful review

- **204-on-malformed.** Malformed / bad-JSON / unknown-action / unknown-resource deliveries are recorded-and-dropped with a **204**, not GitHub's 400 — sustained non-2xx can degrade or disable the webhook on Sentry's side. Silent drops are warn-logged only (no metric); this was an explicit decision.
- **Stable migration table names (deploy step required).** Module migration-tracking tables were named by array index, so a conditionally-composed provider's table name shifted with flag order. We now pin stable names. **Existing environments with GitHub enabled must rename** `__drizzle_migrations_integrations_1` → `__drizzle_migrations_integrations_github` as part of this release, or GitHub migrations re-run against existing tables.
- **`org_slug NOT NULL`** is intentional: no PR1 path inserts an installation row in production (only PR2's connect flow does, and the install redirect always carries `orgSlug`).
- **Signature header fallback.** Accepts both `Sentry-Hook-Signature` and the legacy `sentry-app-signature`, logging which was used — to confirm on first staging delivery.

## Notes for reviewers

- The sentry `drizzle/0000_initial.sql` was hand-authored to mirror the GitHub provider (drizzle's runtime migrator needs only `meta/_journal.json`); regenerate with `drizzle-kit generate` if you prefer generated provenance.
- Issue webhooks must be enabled in the Sentry app config (app-registration concern).

This PR went through `/plan-eng-review` (architecture, tests, performance) and an independent Codex review; the plan and review trail live in `.claude/plans/`.

---
## Summary by cubic
Adds a Sentry provider that ingests Sentry App webhooks, verifies HMAC-SHA256 signatures, dedupes by Request-ID, and publishes normalized `issue.<action>` events to trigger workflows. Internally, the webhook is split into core/presentation modules and drop conditions are modeled as typed errors (no behavior change).

- **New Features**
  - New `@shipfox/api-integration-sentry` + `@shipfox/api-integration-sentry-dto`: webhook receiver with HMAC-SHA256 verification (`Sentry-Hook-Signature` or legacy `sentry-app-signature`), Request-ID dedupe, issue normalization, and `integrations.event.received` publishing with `source: 'sentry'` and `event: issue.<action>`; maps `ignored` → `archived`; malformed/unknown inputs are recorded-and-dropped with 204; schema accepts explicit nulls; DTO schema tests run in CI.
  - Installation lifecycle: `deleted` marks the installation deleted and disables its connection; `created` is recorded only.
  - `@shipfox/api-integration-core`: behind `INTEGRATIONS_ENABLE_SENTRY_PROVIDER`; adds `updateIntegrationConnectionLifecycleStatus`; pins stable per-provider migration table names.
  - `@shipfox/api-integration-core-dto`: adds `SentryIssuePayload`, shared provider-core port types, and exports `SENTRY_ISSUE_ACTIONS`; `@shipfox/api-integration-sentry-dto` derives its action schema from this tuple to prevent drift.
  - `@shipfox/node-fastify`: adds `rawBodyPlugin` and `WEBHOOK_BODY_LIMIT`; `@shipfox/api-integration-github` now uses it and the shared port types (no behavior change).
  - `@shipfox/node-module`: supports `migrationsTableName` for stable migration tracking.

- **Migration**
  - Rename `__drizzle_migrations_integrations_1` → `__drizzle_migrations_integrations_github` before deploy to avoid re-running GitHub migrations.
  - Enable Sentry: set `INTEGRATIONS_ENABLE_SENTRY_PROVIDER=true` and provide `SENTRY_APP_CLIENT_ID`, `SENTRY_APP_CLIENT_SECRET` (used to verify webhook signatures), `SENTRY_APP_SLUG` (optional: `SENTRY_APP_VERIFY_INSTALL`).

<sup>Written for commit c68257419794eb127d5fb7ed00bb14ea3f8379da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libs/api/integration/core/src/index.ts
```diff
@@ -195,7 +237,11 @@ export async function createIntegrationsContext(
 
   const module: ShipfoxModule = {
     name: 'integrations',
-    database: github ? [{db, migrationsPath}, github.database] : {db, migrationsPath},
+    database: [
+      {db, migrationsPath},
+      ...(github ? [{...github.database, migrationsTableName: GITHUB_MIGRATIONS_TABLE}] : []),
```
ThierryAbalea, line 242: This changes GitHub from the old positional migration tracker `__drizzle_migrations_integrations_1` to `__drizzle_migrations_integrations_github`. The changeset documents the required manual rename, but missing that step would make GitHub migrations look unapplied and can block startup by re-running them against existing tables.

Could we make this cutover idempotent in code, for example by renaming the old tracker table before module migrations when it exists? That would keep the stable table names without making deploy correctness depend on a manual DB step.

libs/api/integration/sentry/src/presentation/routes/webhooks.ts
```diff
@@ -0,0 +1,247 @@
...
+
+export function createSentryWebhookRoutes(options: CreateSentryWebhookRoutesOptions): RouteGroup {
+  const webhookRoute = defineRoute({
+    method: 'POST',
+    path: '/',
+    auth: [],
+    description: 'Sentry integration webhook receiver.',
+    options: {bodyLimit: WEBHOOK_BODY_LIMIT},
+    handler: async (request, reply) => {
+      const deliveryId = request.headers[REQUEST_ID_HEADER];
```
ThierryAbalea: This dedupes Sentry deliveries on `Request-ID`. That is fine if Sentry guarantees the same `Request-ID` across retries, but the docs I found only describe it as a tracking/debugging identifier, not as a stable delivery id.

Could we either link/document that retry-stability guarantee here, or use a payload-derived idempotency key? Otherwise a timeout after we commit the outbox event could be retried with a new `Request-ID` and publish the same issue event twice.

libs/api/integration/sentry/src/config.ts
```diff
@@ -0,0 +1,9 @@
+import {bool, createConfig, str} from '@shipfox/config';
+
+export const config = createConfig({
+  SENTRY_APP_CLIENT_ID: str(),
```
ThierryAbalea, line 7: Small config-scope note: this PR only uses `SENTRY_APP_CLIENT_SECRET` for webhook signature verification, but enabling the provider also requires `SENTRY_APP_CLIENT_ID` and `SENTRY_APP_SLUG` at config load. If those are intentionally required now for the next PR's install flow, a short comment would help. Otherwise we could defer requiring the unused values until they are actually consumed.

ThierryAbalea (approved):
Approved with a few non-blocking comments around deploy safety and webhook idempotency.

ThierryAbalea (dismissed):

libs/api/integration/core-dto/src/events.ts
```diff
@@ -18,6 +18,24 @@ export interface GithubPushPayload {
...
+  status: string | null;
+  platform: string | null;
+  webUrl: string | null;
+  issueUrl: string | null;
+  projectUrl: string | null;
+  firstSeenAt: string | null;
+  lastSeenAt: string | null;
+}
+
+export type SentryIssueAction = 'created' | 'resolved' | 'assigned' | 'archived' | 'unresolved';
```
ThierryAbalea: `SentryIssueAction` is defined here and also by the Sentry webhook Zod schema in `@shipfox/api-integration-sentry-dto`. Could we make one source of truth, or add a type-level equivalence check, so adding/removing a Sentry action cannot drift between accepted webhook actions and the published `SentryIssuePayload` contract?

ThierryAbalea (dismissed):
Approved with the existing non-blocking comments plus one note on keeping Sentry issue action contracts in sync.

libs/api/integration/core/src/index.ts
noe-charmet: The GitHub provider isn't enabled in any deployed environment yet, so there is no legacy `__drizzle_migrations_integrations_1` table to rename. This release is the first to create the per-provider trackers at all, so there's no manual step to miss today.

That said, you're right that depending on a manual DB step is fragile for the day it does ship. I'm happy to add an idempotent rename-if-exists guard before module migrations (renaming the old positional tracker when present) so the cutover is safe with no manual step. Say the word and I'll fold it in.

libs/api/integration/sentry/src/presentation/routes/webhooks.ts
noe-charmet: Agreed, this is the weak spot. Same-`Request-ID` retries are safe today because the delivery row insert and the outbox write share one transaction, so a retry with the same id is a no-op. But you're right that I haven't confirmed Sentry keeps `Request-ID` stable across retries, and if it doesn't, a post-commit timeout retried under a new id would double-publish.

I'd rather not depend on the header, so I'll switch the idempotency key to a payload-derived hash (installation uuid + issue id + action + lastSeen) in a follow-up, keeping this PR scoped. Tracking it separately.

libs/api/integration/sentry/src/config.ts
noe-charmet: Intentional. `SENTRY_APP_CLIENT_ID` and `SENTRY_APP_SLUG` are consumed by the install/OAuth + app-token-exchange flow in the next PR; only the signing secret is used by this webhook PR. I'd rather declare the full app credential contract once than toggle requireds across PRs, so I added a comment in `config.ts` (92a9556) noting they're for the upcoming connect flow. Let me know if you'd prefer them deferred until consumed instead.

libs/api/integration/core-dto/src/events.ts
noe-charmet: Good catch. `sentry-dto` already depends on `core-dto`, so I made `core-dto` the single source: it now exports a `SENTRY_ISSUE_ACTIONS` tuple, derives `SentryIssueAction` from it, and `sentry-dto` builds `sentryIssueActionSchema` from the same tuple (92a9556). The accepted webhook actions and the published `SentryIssuePayload` contract can no longer drift.

libs/api/integration/sentry/README.md
```diff
@@ -0,0 +1,79 @@
...
+Configure the Sentry App credentials in the API environment:
+
+```sh
+SENTRY_APP_CLIENT_ID=
+SENTRY_APP_CLIENT_SECRET=
+SENTRY_APP_SLUG=
+SENTRY_APP_VERIFY_INSTALL=true
+```
+
+`SENTRY_APP_CLIENT_SECRET` signs inbound webhooks with HMAC-SHA256.
```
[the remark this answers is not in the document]
noe-charmet: Fixed in c682574. Corrected the README (and the matching `config.ts` comment, which had the same wording) to say the secret is used to verify inbound webhook HMAC-SHA256 signatures: Sentry signs deliveries with the shared secret and the provider verifies them.

libs/api/integration/sentry/src/core/webhook.ts
```diff
@@ -0,0 +1,134 @@
...
+  action: 'created' | 'deleted';
+  installationUuid: string;
+  recordDeliveryOnly: RecordDeliveryOnlyFn;
+  updateConnectionLifecycleStatus: UpdateIntegrationConnectionLifecycleStatusFn;
+}
+
+export async function handleSentryInstallationLifecycle(
+  params: HandleSentryInstallationLifecycleParams,
+): Promise<void> {
+  if (params.action === 'deleted') {
```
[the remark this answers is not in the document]
noe-charmet: Valid structural point. Today the lifecycle side effects are idempotent value-sets (mark the installation `deleted`, set the connection `disabled`), so a duplicate in the normal case just re-applies the same state. The real risk is a late duplicate arriving after a re-install, which would re-disable a freshly re-activated connection.

The right fix is to apply dedup before the side effects, but that depends on the same idempotency-key rework discussed in https://github.com/ShipfoxHQ/shipfox/pull/147#discussion_r3387905075: `Request-ID` may not be stable across Sentry retries, so we want to gate on a payload-derived key, not the header. I'm folding the dedup-before-mutation ordering into that follow-up so a stable key and correct ordering land together.

PR #148, opened 2026-06-10
[api/runners-dto] Add checkout intent + checkout-token DTO schemas

noe-charmet:
Extends `@shipfox/api-runners-dto` with the shared, credential-free contract for runner-managed repository checkouts. Adds a nullable `checkout` intent on the runner job payload and a new checkout-token response DTO.

## Why

The "Runner-managed workspaces" project makes job execution repository-aware: a runner checks out the project repo into a controlled CWD before steps run. The backend will send a typed, **credential-free** checkout intent in the queued job payload, and after a runner claims a job it exchanges its lease for short-lived credentials via a separate endpoint — so usable repository credentials are never stored in queued job rows. This PR lands that contract first (DTO-only, no behavior) so the enqueue enrichment (ENG-407) and the checkout-token endpoint (ENG-406) can build on it in parallel.

## What's in here

- `request-job.ts` — `checkoutIntentSchema` (`repository_url`, `ref`, `provider`, `source_connection_id`, `external_repository_id`) attached to `jobPayloadSchema` as `checkout: …nullish()`.
- `checkout-token.ts` (new) — `checkoutTokenResponseSchema` with a basic/bearer discriminated-union `auth` block and an ISO-datetime-validated `expires_at`.
- First Vitest setup for the package + parse/serialization tests (11 cases).

## Notes for reviewers

- **`checkout` is optional + nullable** on purpose. The `request-job` route is intentionally left untouched — its current response (no `checkout` key) still validates, and old runners ignore the field. The runners suite (76 tests, incl. the unchanged `request-job.test.ts`) stays green, proving backward compatibility.
- **`auth` uses `z.discriminatedUnion('kind', …)`** rather than a single object + `.refine()`, so `kind === 'basic'` narrows `username` to a guaranteed `string` for downstream consumers (ENG-406 / the runner).
- The queued payload deliberately carries **no** token material (the project's rejected shortcut) — credentials are minted post-claim against a job lease.

Refs ENG-402

---
## Summary by cubic
Adds credential-free checkout intent and checkout-token DTOs for runner-managed repo checkouts, now exported from `@shipfox/api-workflows-dto` (not `@shipfox/api-runners-dto`). DTO-only, backward compatible, aligned with ENG-402.

- New Features
  - Exported `checkoutIntentSchema` (standalone) for backend checkout resolution.
  - Added `checkoutTokenResponseSchema` with `auth` `'basic' | 'bearer'` and ISO `expires_at` (offset allowed); all intent/auth strings are non-empty.

- Refactors
  - Relocated checkout DTOs to `@shipfox/api-workflows-dto` (Orchestration owns checkout); runners DTOs remain checkout-agnostic.
  - Added Vitest setup and schema tests in `@shipfox/api-workflows-dto`; added a `jobPayload` validation test in `@shipfox/api-runners-dto`.

<sup>Written for commit 3bc02de07b85dd86a2ff444b90dc4265180aa9f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libs/api/runners-dto/src/schemas/request-job.ts
```diff
@@ -10,11 +10,22 @@ export const jobPayloadStepSchema = z.object({
...
+});
+
+export type CheckoutIntentDto = z.infer<typeof checkoutIntentSchema>;
+
 export const jobPayloadSchema = z.object({
   job_id: z.string().uuid(),
   run_id: z.string().uuid(),
   job_name: z.string(),
   steps: z.array(jobPayloadStepSchema).min(1),
+  checkout: checkoutIntentSchema.nullish(),
```
ThierryAbalea: This adds `checkout` to `JobPayloadDto`, but #143 moves runner scheduling/claiming to the step-less protocol: pending rows store only `{job_id, run_id, workspace_id}` and `/request` returns `{job_id, run_id, lease_token}`, not the job payload. If #143 lands first, this checkout intent will be attached to a DTO the new runner protocol no longer delivers. Should we rebase this on the post-#143 contract and put the checkout intent/metadata on the future lease-scoped checkout endpoint instead?

libs/api/runners-dto/src/schemas/checkout-token.ts
```diff
@@ -0,0 +1,29 @@
+import {z} from 'zod';
+
+const checkoutTokenBasicAuthSchema = z.object({
+  kind: z.literal('basic'),
+  username: z.string(),
+  token: z.string(),
```
ThierryAbalea: Could we use `.min(1)` for the credential fields, and probably the checkout intent strings too? Empty `token`, `username`, `repository_url`, `ref`, `provider`, or `external_repository_id` values are not useful, and allowing them means a malformed checkout payload passes DTO validation before failing later in runner checkout logic.

libs/api/runners-dto/src/schemas/checkout-token.ts
```diff
@@ -0,0 +1,29 @@
...
+  expires_at: z.string().datetime(),
```
ThierryAbalea: `z.string().datetime()` only accepts the `...Z` form by default and rejects offset timestamps like `2026-06-10T12:00:00+00:00`. If the checkout-token issuer will always use `toISOString()`, this is fine; otherwise could we use `.datetime({offset: true})` or add a test that makes the intended wire format explicit?

ThierryAbalea (dismissed):
Approved with three non-blocking comments on the checkout contract alignment and schema strictness.

libs/api/runners-dto/src/schemas/request-job.ts
noe-charmet: Good catch — confirmed: after #143/#145 `jobPayloadSchema` is orphaned (`/request` returns `{job_id, run_id, lease_token}`, steps come from `/runs/jobs/current/steps/next`). Dropped the `checkout` field from the payload in be66dbf and kept `checkoutIntentSchema` as a standalone exported type the backend resolves at checkout time. The runner-facing contract is now `checkoutTokenResponseSchema` (lease-scoped), which already carries `repository_url`+`ref`+credentials. Delivery wiring stays with ENG-406; ENG-407 (enqueue enrichment) will be re-scoped onto the step-less contract.

libs/api/runners-dto/src/schemas/checkout-token.ts
noe-charmet: Done in be66dbf — added `.min(1)` to the credential fields (`token`, `username`) and the checkout-token `repository_url`/`ref`, plus the `checkoutIntentSchema` strings (`repository_url`, `ref`, `provider`, `external_repository_id`). Added empty-string rejection tests.

libs/api/runners-dto/src/schemas/checkout-token.ts
noe-charmet: Switched to `.datetime({offset: true})` in be66dbf so both the `toISOString()` Z form and offset timestamps parse. Added two tests pinning the wire format explicit: one for the `...Z` form and one for a `+02:00` offset.

PR #149, opened 2026-06-10
[api/integration] Add createCheckoutSpec() for GitHub and Debug

noe-charmet:
Exposes `createCheckoutSpec()` on `IntegrationSourceControlService` so the backend can resolve checkout data and credential material from the source-control provider and hand a runner a typed, credential-bearing checkout intent. Part of the Runner-managed workspaces project; this PR settles the service contract that the checkout-token endpoint and runner enrichment build on.

## What changed

- **core-dto** — `CheckoutSpec` gains an optional structured `credentials` field (`CheckoutCredentials { username, token, expiresAt }`); `CreateCheckoutSpecInput.ref` is now optional. Adds a pure `redactCheckoutSpec()` helper that masks the token for logging. Wires the package for unit tests (it had none).
- **core** — `createCheckoutSpec()` on the service, following the existing dispatch pattern (connection + workspace guards). New typed `IntegrationCheckoutUnsupportedError` for providers that don't implement the optional method.
- **github** — `createInstallationAccessToken()` on the API client (repo-scoped, `contents: read`); the provider returns a clean `repositoryUrl` plus `credentials { username: 'x-access-token', token, expiresAt }`, defaulting `ref` to the repository default branch.
- **debug** — returns the static clone URL with no credentials.

## Why credentials are a structured field, not embedded in the URL

An inline-token URL (`https://x-access-token:<token>@github.com/...`) is the leakiest representation — it ends up in `git remote -v`, `git clone <url>` process args, and any log/row that serializes the spec, which conflicts with the project's hard "no credentials in logs, API logs, queued/running job rows, or persisted results" criterion. A structured field makes redaction trivial, keeps `repositoryUrl` secret-free, and conveys token expiry. (This decision was made in plan review and is documented on the Linear issue.)

## Notes for reviewers

- This is purely additive and **dormant**: no HTTP route consumes it yet. `redactCheckoutSpec()` has no call site in this PR by design — the dependent checkout-token endpoint and runner enrichment tickets wire it in, compose auth from `credentials`, and respect `expiresAt`.
- The GitHub installation token is GitHub-issued (not a Shipfox-signed token), so it stays in the github package rather than api-auth.
- The Octokit call in `client.ts` is not unit-tested, matching the existing posture (no github client method is unit-tested; it's mocked at the provider level).

Refs ENG-403

---
## Summary by cubic
Adds `createCheckoutSpec()` to the source-control service and GitHub/Debug providers to return a clean checkout spec with optional short-lived credentials for runners. Hardens `redactCheckoutSpec()` to also strip any credentials embedded in the `repositoryUrl`. Supports ENG-403; dormant until the checkout-token endpoint and runner enrichment are wired in.

- **New Features**
  - Service: `createCheckoutSpec()` returns a typed `CheckoutSpec` with optional `credentials`; enforces workspace/connection guards; unsupported providers throw `IntegrationCheckoutUnsupportedError`.
  - GitHub: mints a short‑lived, repo‑scoped installation token (`contents: read`), validates a parseable expiry, classifies malformed responses as `malformed-provider-response`, and returns a clean `repositoryUrl` plus `credentials { username: 'x-access-token', token, expiresAt }`; `ref` defaults to the repository’s default branch; adds a client test pinning `repository_ids` scoping and permissions.
  - Debug: returns the static clone URL with no credentials; `ref` defaults to the repository’s default branch.
  - DTO (`@shipfox/api-integration-core-dto`): adds `CheckoutCredentials`, makes `CreateCheckoutSpecInput.ref` optional, adds `redactCheckoutSpec()` to mask tokens and strip URL userinfo as defense in depth, and documents that credentials must never be embedded in `repositoryUrl`; adds test scripts/config and unit tests.
  - Core (`@shipfox/api-integration-core`): exports `CheckoutSpec`, `CheckoutCredentials`, `CreateCheckoutSpecInput`, `redactCheckoutSpec`, and `IntegrationCheckoutUnsupportedError` from the package root for public use.
  - Tests: coverage for redaction (including URL-embedded creds) and provider/service flows across `@shipfox/api-integration-core-dto`, `@shipfox/api-integration-core`, `@shipfox/api-integration-github`, and `@shipfox/api-integration-debug`, plus a GitHub client test validating token scope and expiry.

<sup>Written for commit 67eb6f74201874e74e9795638690f01cb69b4ef2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/149?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libs/api/integration/core-dto/src/contracts/integrations.ts
+++ b/libs/api/integration/core-dto/src/contracts/integrations.ts
@@ -167,11 +167,31 @@ export const MAX_REPOSITORY_FILE_BYTES = 1_000_000;
 const REDACTED_TOKEN = '***';
 
 export function redactCheckoutSpec(spec: CheckoutSpec): CheckoutSpec {
-  if (!spec.credentials) return spec;
+  const repositoryUrl = stripUrlCredentials(spec.repositoryUrl);
+  if (!spec.credentials) {
+    return repositoryUrl === spec.repositoryUrl ? spec : {...spec, repositoryUrl};
+  }
   return {
     ...spec,
+    repositoryUrl,
     credentials: {...spec.credentials, token: REDACTED_TOKEN},
   };
+}
+
+// Defense in depth: providers must keep `repositoryUrl` credential-free, but a
+// helper whose job is to make a spec safe to log must also strip any userinfo
+// (e.g. `https://x-access-token:<token>@host/...`) so a provider mistake cannot
+// leak a token through the URL.
+function stripUrlCredentials(repositoryUrl: string): string {
+  try {
+    const url = new URL(repositoryUrl);
+    if (!url.username && !url.password) return repositoryUrl;
+    url.username = '';
+    url.password = '';
+    return url.toString();
+  } catch {
+    return repositoryUrl;
+  }
 }
 
 export function buildProviderRepositoryId(
--- a/libs/api/integration/core-dto/src/contracts/integrations.ts
+++ b/libs/api/integration/core-dto/src/contracts/integrations.ts
@@ -87,6 +87,12 @@ export interface CheckoutCredentials {
 }
 
 export interface CheckoutSpec {
+  /**
+   * Clone URL that must never embed authentication material. Credentials live
+   * only in `credentials` so `redactCheckoutSpec()` can mask them; a provider
+   * that embeds a token in this URL would bypass redaction and leak it into
+   * logs, `git remote -v`, and persisted job rows.
+   */
   repositoryUrl: string;
   ref: string;
   credentials?: CheckoutCredentials | undefined;
--- a/libs/api/integration/core-dto/src/contracts/integrations.ts
+++ b/libs/api/integration/core-dto/src/contracts/integrations.ts
@@ -80,15 +80,22 @@ export interface FetchFileInput<Connection extends IntegrationConnection = Integ
   path: string;
 }
 
+export interface CheckoutCredentials {
+  username: string;
+  token: string;
+  expiresAt: Date;
+}
+
 export interface CheckoutSpec {
   repositoryUrl: string;
   ref: string;
+  credentials?: CheckoutCredentials | undefined;
 }
 
 export interface CreateCheckoutSpecInput<
   Connection extends IntegrationConnection = IntegrationConnection,
 > extends ResolveRepositoryInput<Connection> {
-  ref: string;
+  ref?: string | undefined;
 }
 
 export interface SourceControlProvider<
@@ -150,6 +157,16 @@ export class IntegrationProviderError extends Error {
 }
 
 export const MAX_REPOSITORY_FILE_BYTES = 1_000_000;
+
+const REDACTED_TOKEN = '***';
+
+export function redactCheckoutSpec(spec: CheckoutSpec): CheckoutSpec {
+  if (!spec.credentials) return spec;
+  return {
+    ...spec,
+    credentials: {...spec.credentials, token: REDACTED_TOKEN},
+  };
+}
 
 export function buildProviderRepositoryId(
   provider: IntegrationProviderKind,
--- a/libs/api/integration/github/src/api/client.ts
+++ b/libs/api/integration/github/src/api/client.ts
@@ -88,6 +88,15 @@ export interface GithubApiClient {
     ref: string;
     path: string;
   }): Promise<GithubFileContent>;
+  createInstallationAccessToken(input: {
+    installationId: number;
+    repositoryId: number;
+  }): Promise<GithubInstallationAccessToken>;
+}
+
+export interface GithubInstallationAccessToken {
+  token: string;
+  expiresAt: Date;
 }
 
 export function createGithubApiClient(): GithubApiClient {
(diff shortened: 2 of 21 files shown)


libs/api/integration/core-dto/src/contracts/integrations.ts
```diff
@@ -80,15 +80,22 @@ export interface FetchFileInput<Connection extends IntegrationConnection = Integ
...
 }
 
+export interface CheckoutCredentials {
+  username: string;
+  token: string;
+  expiresAt: Date;
+}
+
 export interface CheckoutSpec {
   repositoryUrl: string;
```
ThierryAbalea, line 96: `redactCheckoutSpec()` only masks `credentials.token`, so the security model depends on every provider keeping `repositoryUrl` credential-free. GitHub and Debug do that today, but the DTO does not state the invariant.

Could we document on `CheckoutSpec.repositoryUrl` that credentials must never be embedded there and must live only in `credentials`? That should help future providers avoid bypassing redaction by accident.

libs/api/integration/github/src/api/client.ts
```diff
@@ -329,6 +338,31 @@ class OctokitGithubApiClient implements GithubApiClient {
     };
   }
 
+  async createInstallationAccessToken(input: {
+    installationId: number;
+    repositoryId: number;
+  }): Promise<GithubInstallationAccessToken> {
+    const response = await mapGithubError(() =>
+      this.getApp().octokit.rest.apps.createInstallationAccessToken({
```
ThierryAbalea, line 346: This is the security-critical bit that keeps the runner token least-privilege: one repository and `contents: read`. The provider test only asserts the mocked client was called with `{ installationId, repositoryId }`, so a later client refactor could drop `repository_ids` or widen permissions without a test failing.

Could we add a small client-level test that pins the Octokit call to `repository_ids: [repositoryId]` and `permissions: { contents: 'read' }`?

libs/api/integration/github/src/api/client.ts
```diff
@@ -329,6 +338,31 @@ class OctokitGithubApiClient implements GithubApiClient {
...
+    if (typeof response.data.token !== 'string') {
+      throw new GithubIntegrationProviderError(
+        'malformed-provider-response',
+        'GitHub installation access token response did not include a token',
+      );
+    }
+
+    return {
+      token: response.data.token,
+      expiresAt: new Date(response.data.expires_at),
```
ThierryAbalea: We validate that GitHub returned a token, but not that `expires_at` is present and parseable. If GitHub or Octokit ever returns a malformed response here, `new Date(undefined)` or another invalid date would flow into `CheckoutCredentials.expiresAt` instead of being classified as `malformed-provider-response`.

Could we validate `response.data.expires_at` before returning the checkout credentials?

ThierryAbalea (commented):
Left three non-blocking comments on the credential-bearing checkout spec contract and GitHub token handling.

ThierryAbalea (dismissed):
Approved with the three non-blocking comments already left on the checkout credential contract and GitHub token handling.

libs/api/integration/core-dto/src/contracts/integrations.ts
noe-charmet: Done in 93e2b4d. Added a doc comment on `CheckoutSpec.repositoryUrl` stating credentials must never be embedded there and must live only in `credentials`, so a future provider can't bypass `redactCheckoutSpec()` by accident.

libs/api/integration/github/src/api/client.ts
noe-charmet: Done in 93e2b4d. Added `src/api/client.test.ts` with a scoped `octokit` mock that pins the call to `repository_ids: [repositoryId]` and `permissions: { contents: 'read' }`, so a later client refactor that drops the repo scope or widens permissions now fails the test.

libs/api/integration/github/src/api/client.ts
noe-charmet: Done in 93e2b4d. `createInstallationAccessToken` now validates `expires_at` (`new Date(...)` rejected when `NaN`) and throws `malformed-provider-response` instead of letting an invalid Date reach `CheckoutCredentials.expiresAt`. Covered by a new client test case.

libs/api/integration/core-dto/src/contracts/integrations.ts
```diff
@@ -151,6 +164,16 @@ export class IntegrationProviderError extends Error {
...
 export const MAX_REPOSITORY_FILE_BYTES = 1_000_000;
 
+const REDACTED_TOKEN = '***';
+
+export function redactCheckoutSpec(spec: CheckoutSpec): CheckoutSpec {
+  if (!spec.credentials) return spec;
+  return {
+    ...spec,
+    credentials: {...spec.credentials, token: REDACTED_TOKEN},
+  };
```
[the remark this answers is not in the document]
noe-charmet: Good catch — fixed in 67eb6f7. `redactCheckoutSpec()` now also strips userinfo from `repositoryUrl` via `URL` (with a fallback to the original string for non-URL values), so a provider that mistakenly embeds a token in the URL can't leak it through logs. It's defense in depth on top of the `repositoryUrl`-must-be-credential-free invariant. Added two tests: a tokenized clone URL is stripped both with and without a `credentials` field present.

PR #150, opened 2026-06-10
Add per-step runner execution protocol

ThierryAbalea:
Backend contract for per-step execution on top of the runner step endpoints from #145.

What changes:
- Adds the workflow-host `nextStep` / `reportStep` path needed by a future runner consumer.
- Step reports carry `exit_code` and an optional `attempt` on the wire.
- Job completion is derived once from terminal step state and emitted through `WORKFLOWS_JOB_COMPLETED` in the same transaction as the final step result.
- The existing legacy completion path remains compatible during rollout.

This PR intentionally does **not** change the runner execution loop. Runner consumption of the per-step protocol is deferred so this PR only establishes the backend contract and durable completion boundary.

Base note: this PR is based on the temporary `durable-gate/145-rebased` shim. After #145 lands on `main`, this branch must be manually rebased onto `main` by dropping the shim commits, then descendants can use the normal stack flow.

Refs ENG-397

## Stack

Part 1 of 6.

Depends on: none
Next: #152

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #150 | `durable-gate/01-per-step-protocol` | `main` |
| 2 | #152 | `durable-gate/02-step-attempts` | `durable-gate/01-per-step-protocol` |
| 3 | #153 | `durable-gate/03-pure-decision` | `durable-gate/02-step-attempts` |
| 4 | #154 | `durable-gate/04-gate-eval-terminal` | `durable-gate/03-pure-decision` |
| 5 | #155 | `durable-gate/05-durable-restart` | `durable-gate/04-gate-eval-terminal` |
| 6 | #156 | `durable-gate/06-dashboard-attempts` | `durable-gate/05-durable-restart` |

### DIFF
--- a/.changeset/brisk-falcons-relay.md
+++ b/.changeset/brisk-falcons-relay.md
@@ -5,4 +5,4 @@
 "@shipfox/api-runners-dto": minor
 ---
 
-Adds per-step runner execution: the runner pulls and reports steps over the lease-token-authed step API (minted at job claim), step reports carry exit codes, and job completion is signalled through the workflows outbox instead of a single atomic job report.
+Adds the backend contract for per-step execution: job claims mint a lease token for the step API, step reports carry attempts and exit codes, and workflow completion can be signalled through the workflows outbox. The runner-side step loop is intentionally owned by the follow-up runner protocol work.
--- a/.changeset/brisk-falcons-relay.md
+++ b/.changeset/brisk-falcons-relay.md
@@ -1,0 +1,8 @@
+---
+"@shipfox/api-workflows": minor
+"@shipfox/api-workflows-dto": minor
+"@shipfox/api-runners": minor
+"@shipfox/api-runners-dto": minor
+---
+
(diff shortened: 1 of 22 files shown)


noe-charmet:
I guess this is the PR that mostly overlaps with ongoing work on the [runner step protocol](https://linear.app/shipfox/project/per-step-runner-protocol-bef2f6a5d683/issues?layout=list&ordering=priority&grouping=workflowState&subGrouping=none&showCompletedIssues=all&showSubIssues=true&showTriageIssues=true). 
Looks like this PR attempts to both handle the [agent step loop](https://linear.app/shipfox/issue/ENG-399/agent-per-step-loop) who has a [Related PR](https://github.com/ShipfoxHQ/shipfox/pull/163)
As well as the finalization which would be [addressed in this ticket](https://linear.app/shipfox/issue/ENG-400/orchestration-per-step-cutover-remove-atomic-path)

PR #152, opened 2026-06-10
Track per-step execution attempts

ThierryAbalea:
Adds durable per-step attempt history.

What changes:
- Adds `workflows_step_attempts` and `steps.current_attempt`.
- `nextStepForJob` opens a running attempt when dispatching a step.
- `recordStepResult` finalizes the current attempt and is attempt-aware: duplicate final reports are idempotent, stale older reports are ignored, and ahead-of-current reports are rejected.
- `workflows_step_attempts.exit_code` is the canonical persisted exit code for attempt history; report output is also persisted on the attempt row.
- DB checks enforce positive attempt numbers and no `pending` attempt rows.
- Migration boundary: steps already running before this table exists may not have a dispatch-time attempt row, so reporting backfills a running attempt row before finalization. Its `started_at` is approximate in that case.

Refs ENG-397

## Stack

Part 1 of 5.

Depends on: none
Next: #153

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #152 | `durable-gate/02-step-attempts` | `main` |
| 2 | #153 | `durable-gate/03-pure-decision` | `durable-gate/02-step-attempts` |
| 3 | #154 | `durable-gate/04-gate-eval-terminal` | `durable-gate/03-pure-decision` |
| 4 | #155 | `durable-gate/05-durable-restart` | `durable-gate/04-gate-eval-terminal` |
| 5 | #156 | `durable-gate/06-dashboard-attempts` | `durable-gate/05-durable-restart` |

### DIFF
--- a/libs/api/workflows-dto/src/schemas/job-execution.ts
+++ b/libs/api/workflows-dto/src/schemas/job-execution.ts
@@ -9,9 +9,9 @@ export const nextStepResponseSchema = z.discriminatedUnion('kind', [
   z.object({
     kind: z.literal('step'),
     step: stepDtoSchema,
-    // The attempt number this dispatch runs. The runner echoes it back on report
-    // so a late report from a superseded attempt can be ignored. Always 1 until
-    // per-step attempts land (PR B: steps.current_attempt).
+    // The attempt number this dispatch runs (the step's current attempt). The
+    // runner echoes it back on report so a late report from a superseded attempt
+    // is ignored. 1 until a durable restart (PR E) bumps it.
     attempt: z.number().int().positive(),
   }),
   z.object({
@@ -27,12 +27,12 @@ export const reportStepBodySchema = z
     status: z.enum(['succeeded', 'failed']),
     error: stepErrorDtoSchema.optional(),
     // The attempt the runner was dispatched, echoed from next-step. Optional on
-    // the wire for now; idempotency enforcement against it lands in PR B.
+    // the wire (older runners omit it); when present it drives attempt-aware
+    // idempotency in `recordStepResult`.
     attempt: z.number().int().positive().optional(),
-    // Process exit code on success and failure. The runner sends it, but it is
-    // not persisted yet: PR B stores it per attempt and PR D consumes it for
-    // gates (success_if: exit_code == 0). On failure it is also carried under
-    // `error.exit_code`.
+    // Process exit code on success and failure. Persisted on the step attempt;
+    // consumed by gate evaluation (success_if: exit_code == 0) in PR D. On
+    // failure it is also carried under `error.exit_code`.
     exit_code: z.number().int().nullable().optional(),
   })
   .refine((body) => (body.status === 'succeeded' ? body.error == null : body.error != null), {
--- a/libs/api/workflows-dto/src/schemas/job-execution.ts
+++ b/libs/api/workflows-dto/src/schemas/job-execution.ts
@@ -9,14 +9,19 @@ export const nextStepResponseSchema = z.discriminatedUnion('kind', [
   z.object({
     kind: z.literal('step'),
     step: stepDtoSchema,
-    // The attempt number this dispatch runs (the step's current attempt). The
-    // runner echoes it back on report so a late report from a superseded attempt
-    // is ignored. 1 until a durable restart (PR E) bumps it.
-    attempt: z.number().int().positive(),
+    attempt: z
+      .number()
+      .int()
+      .positive()
+      .describe(
+        'Attempt number for this step dispatch. Echo it back when reporting the step result so stale reports from older attempts can be ignored.',
+      ),
   }),
   z.object({
     kind: z.literal('done'),
-    status: z.enum(['succeeded', 'failed']),
+    status: z
+      .enum(['succeeded', 'failed'])
+      .describe('Terminal status of the job when no step remains to run.'),
   }),
 ]);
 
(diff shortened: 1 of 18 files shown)


noe-charmet:
This is overlapping and contradicting with the work we're doing on moving all runner orchestration to the worker module.
[Typically this PR](https://github.com/ShipfoxHQ/shipfox/pull/145)

I believe this concern of tracking step state should only be in the worklow module and not at all in the runner module like it is proposed here.

noe-charmet:
I'm not sure I fully understand, conceptually, why we need both a distinct `StepAttempt` and a `Step` with an `attempt` property.

I kind of follow that `StepAttempt` is meant to represent a single execution view, but why isn't `Step` itself that same thing?

If the idea is that we need some kind of view sitting between the workflow "state" and the execution status history, then I think we should:

- Improve the naming and separation of entities in the code organization. And what about workflow-level retries? Would we also need something like a `WorkflowRun` concept?
- Be extra careful around state duplication and transactionality. My gut feeling is that we're trying to keep some state in the DB while also keeping an event history. Would it be possible to derive the current state from the event history instead? The `currentAttempt` counter, for example, could drift from reality pretty easily if anything happens non-transactionally.

ThierryAbalea:
> overlapping and contradicting with the work we're doing on moving all runner orchestration to the worker module.

I disagree that this moves orchestration into the runner. It moves job-completion derivation from the legacy runner `completeJob` path into the workflow host, while Temporal/worker still owns durable finalization. The runner only controls polling cadence and executes/reports steps.

I think the #152+ direction is better because the backend already has the authoritative persisted step state, so job completion should be derived from that state instead of depending on one extra runner call. It also removes the failure window where the runner successfully reports the last step, then crashes before `completeJob`, leaving the job `running` until timeout.

noe-charmet:
Looks like a UI diff issue with the branch targeting non-main branches. I was seeing new endpoints and HTTP interactions for step fetching/reporting. Looks OK now.
I think we're fully aligned on the way to go with orchestration and step fetching/reporting. No discussion to have on this topic.

libs/api/workflows-dto/src/schemas/job-execution.ts
```diff
@@ -9,9 +9,9 @@ export const nextStepResponseSchema = z.discriminatedUnion('kind', [
   z.object({
     kind: z.literal('step'),
     step: stepDtoSchema,
-    // The attempt number this dispatch runs. The runner echoes it back on report
-    // so a late report from a superseded attempt can be ignored. Always 1 until
-    // per-step attempts land (PR B: steps.current_attempt).
+    // The attempt number this dispatch runs (the step's current attempt). The
```
noe-charmet: Not for this PR specifically, but we should target to land a proper user-facing description for the schema. Zod has a `.describe()` property for this. The inline comments are pretty much useless for any external usage.
Just like we do in [this PR](https://github.com/ShipfoxHQ/shipfox/pull/162) for config

ThierryAbalea:
> I'm not sure I fully understand, conceptually, why we need both a distinct `StepAttempt` and a `Step` with an `attempt` property.
> 
> I kind of follow that `StepAttempt` is meant to represent a single execution view, but why isn't `Step` itself that same thing?
> 
> If the idea is that we need some kind of view sitting between the workflow "state" and the execution status history, then I think we should:
> 
> * Improve the naming and separation of entities in the code organization. And what about workflow-level retries? Would we also need something like a `WorkflowRun` concept?
> * Be extra careful around state duplication and transactionality. My gut feeling is that we're trying to keep some state in the DB while also keeping an event history. Would it be possible to derive the current state from the event history instead? The `currentAttempt` counter, for example, could drift from reality pretty easily if anything happens non-transactionally.

I agree the `Step` / `StepAttempt` split needs to be made explicit.

The intended model is:

- `steps` is the current projection used by scheduling, locking, dashboard/API reads, and current job state.
- `step_attempts` is execution history and an idempotency/stale-report anchor.

So `Step` is not meant to represent one execution attempt. It represents the current state of a logical workflow step in a run. `StepAttempt` represents one concrete execution try of that step.

On deriving current state from history: yes, it is possible in principle, but it would be a significant architecture change. The current Temporal history is mostly about job/run lifecycle transitions. It does not contain the full step execution history as the authoritative runtime state. To rely on Temporal history for this, we would need to move step state transitions into the Temporal workflow itself and make Temporal state authoritative.

Today, the authoritative runtime state is already in Postgres: `workflows_steps`, `workflows_jobs`, and `workflows_workflow_runs`. Those tables are not only read models today; they are the runtime state the system mutates and reasons from. Existing read paths, such as `getWorkflowRunAggregates`, `listRuns`, and `getRun`, already query those same tables.

If we make Temporal authoritative instead, `workflows_steps` could still exist, but it would no longer be the authoritative step-state table. It would become a Postgres projection of Temporal state for dashboard/API reads. That means we would need to design and maintain projection synchronization: Temporal state changes would need to update Postgres, retries/idempotency would need to be handled, and we would need rules for projection lag or rebuild.

That is why I think the better fit for this PR stack is to keep Postgres as the authoritative runtime state and make `step_attempts` the history/audit/idempotency layer. The important invariant is that `steps.current_attempt` must not drift from `step_attempts`: it is a projection pointer, not an independent source of truth. Every mutation of `steps.status` / `steps.current_attempt` and the corresponding `step_attempts` row must happen in the same transaction under the same lock.

I’ll update the PR to make that boundary explicit and add invariant tests around projection/history consistency.

noe-charmet:
Just to clarify my earlier point, I wasn't suggesting we bring in Temporal for state management. My question was more about the current design of storing `workflows_jobs` / `workflows_steps` directly in Postgres.

I'm wondering whether those could instead be derived as event-sourced state from the proposed `step_attempts` table, rather than being stored as separate first-class tables.

This isn't a blocker. I'm mainly trying to understand the reasoning behind the current design choice so we can be confident we're landing on the best approach.

ThierryAbalea:
> Just to clarify my earlier point, I wasn't suggesting we bring in Temporal for state management. My question was more about the current design of storing `workflows_jobs` / `workflows_steps` directly in Postgres.
> 
> I'm wondering whether those could instead be derived as event-sourced state from the proposed `step_attempts` table, rather than being stored as separate first-class tables.
> 
> This isn't a blocker. I'm mainly trying to understand the reasoning behind the current design choice so we can be confident we're landing on the best approach.

Current state shouldn't be derived from attempts.

The blocking reason is the pending graph. Before a step runs it has no attempt, but we still need its row to lock and serialize the step endpoints (`nextStepForJob`/`recordStepResult`) and to know the ordered set of not-yet-run steps. Attempts can't represent a step that hasn't executed.

Two follow-on cases confirm it:

- Cancellation: when step 1 fails, later pending steps become `cancelled` without ever running. On `workflows_steps.status` that's one explicit write; derived state needs fake cancelled attempts or special inference.
- Terminal check: "job done" = all steps `succeeded`/`failed`/`cancelled`, read directly from `workflows_steps.status` instead of recomputed from attempts plus cancellation rules.

So `workflows_steps` holds current state for reads and locking; `workflows_step_attempts` stays pure execution history.

Separately on vocab: let's keep `run` for the top-level context, one definition plus one trigger/input, and `attempt` for repeated tries inside it. `execution` is too broad because it can read at workflow, job, step, or runner level. `StepAttempt` means one try of a step within a `WorkflowRun`, and this extends cleanly to `JobAttempt` if we later retry whole jobs.

ThierryAbalea:
Fixed the schema-description issue from the DTO review: `libs/api/workflows-dto/src/schemas/job-execution.ts` now uses Zod `.describe()` for the public next-step/report-step fields instead of relying on inline comments. The stack was restacked and #152-#156 were force-with-lease pushed.

PR #161, opened 2026-06-11
[api/auth] Document the lease token security perimeter

noe-charmet:
Reworks the auth package's `## Auth methods` README section into a code-free `## Security model` overview covering both bearer tokens, and adds matching guardrails to `AGENTS.md`.

The user session token kept its design/lifecycle/tradeoff shape, abstracted away from secret names and route paths. The job lease token is rebuilt around the six areas ENG-401 asked for, most of which were previously missing or only implied:

- **Scope** — states the negatives: not a runner identity, not workspace-wide, not a substitute for the runner credential.
- **Trust boundary** — single issuer (scheduling), everyone else verify-only in-process with no callback, runner/agent untrusted and only presents the token.
- **Mechanics** — HS256, dedicated secret, fixed audience to block cross-purpose replay, bounded TTL, secret from config, never log the raw token (no auto-redaction safety net).
- **Defense in depth** — server job/step state is the final gate; terminal-state reports are ignored; cancellation rides the heartbeat response.
- **Threat model** — small blast radius for a leaked/replayed token; secret leak → rotate → all live leases invalid.
- **Guidelines for future changes** — don't widen claims, bound the TTL, single issuer, verify-only elsewhere.

The doc stays deliberately abstract from the implementation (no file or function names) — it's a security overview, not a build guide. The existing code comment that referenced the old section name was repointed.

Two gaps surfaced while writing that are intentionally **not** addressed here (docs-only PR), worth follow-ups:
- There is no explicit job-finalized gate in code today — safety comes from step-state idempotency. The doc describes that reality rather than overclaiming.
- No log redaction is configured anywhere, so "never log the token" has no automatic backstop.

Refs ENG-401

---
## Summary by cubic
Reworked `libs/api/auth` into a code-free Security model for both bearer tokens and defined the job lease token’s perimeter (scope, trust boundary, mechanics, defense-in-depth, threat model, change guidelines). Clarifies that the lease path gates on step/progression state only (job finalization is enforced outside it), adds matching guardrails in `AGENTS.md`, and updates the lease auth comment to reference the new section; completes ENG-401.

<sup>Written for commit 9872a72f2e1082efb2e02b2360aab402eaa6f691. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libs/api/auth/README.md
+++ b/libs/api/auth/README.md
@@ -55,62 +55,88 @@ Required environment:
 | `SMTP_USER` | none | Optional SMTP user. |
 | `SMTP_PASSWORD` | none | Optional SMTP password. |
 
-## Auth methods
+## Security model
 
-The module exposes two Fastify auth methods. Both read an HS256 token from the
-`Authorization: Bearer <token>` header and verify it statelessly — signature and
-expiry are checked, but the claims are trusted without a database read. They
-differ in who they authenticate and what they grant access to, which is what
-makes the stateless tradeoff acceptable in each case.
+The module issues two kinds of bearer token, both presented as
+`Authorization: Bearer <token>`. Both are **stateless**: each is signed with
+HMAC-SHA256 and verified by checking its signature and expiry alone, with no
+database read on the request path. They differ in who they authenticate and what
+they grant, and the stateless tradeoff is accepted for a different reason in each
+case. Each is signed with its own dedicated secret, so neither token type can be
+used in place of the other.
 
-### User JWT (`createJwtAuthMethod`)
+### User session token
 
-- **Design:** a stateless session token. Signed with `AUTH_JWT_SECRET`, it carries
-  the user identity and membership list in its claims so protected routes can
-  authorize without a per-request user/membership lookup.
+- **Design:** a stateless session token. It carries the user's identity and
+  current membership list in its claims so protected routes can authorize without
+  a per-request user or membership lookup.
 - **Audience:** signed-in users on first-party clients (web app, CLI).
-- **Grants access to:** the user's own account routes (`/me`, `/change-password`)
-  and, downstream, any route that authorizes against the membership claims. Scope
-  is "this user, with these memberships" — not a single resource.
+- **Grants:** the user's own account routes and, downstream, any route that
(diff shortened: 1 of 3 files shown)


libs/api/auth/README.md
```diff
@@ -55,62 +55,88 @@ Required environment:
...
+  presents the one it was handed.
+- **Mechanics.** Signed with HMAC-SHA256 using a dedicated secret, separate from
+  the user-session secret. Its claims name the job and its surrounding context and
+  nothing more, and it carries a fixed audience so a token minted for one purpose
+  cannot be replayed against another. It is short-lived (bounded in hours, not
+  days). The signing secret is supplied through configuration, never embedded in
+  code or committed. The raw token must **never** be written to logs, traces, or
+  error payloads — there is no automatic redaction to fall back on.
+- **Defense in depth — server state is the final authority.** A valid token is
+  never sufficient on its own to advance work. Server-side job and step state is
```
ThierryAbalea: The security model says "server-side job and step state" is the final gate, but the leased workflow route path appears to enforce this through step-row state only (`recordStepResult` ignores terminal steps; I do not see a job-status check before accepting a leased step report). Could we soften this to "server-side step/progression state," or note that job finalization is handled outside the lease route path?

ThierryAbalea (dismissed):
Reviewed.

libs/api/auth/README.md
noe-charmet: Good catch — you're right that the lease route path gates on step-row/progression state only, not job status. Reworded both the README and AGENTS.md: the lease path now reads as gating on "server-side step and progression state," with a note that job-level finalization is enforced outside the lease path. Fixed in 716124f.

PR #163, opened 2026-06-11
[runner] Switch the agent to the per-step pull/report loop

noe-charmet:
Switches the runner agent from the job-atomic protocol to the per-step pull/report loop. The runner claims a job, takes the lease token from the claim response, then loops `next -> execute -> report` against the orchestration step API until the job is done. The atomic `executeJob`/`completeJob` path and `executor.ts` are removed.

This builds on the per-step **backend contract already merged to `main`** ("Add per-step runner execution protocol"), and is the runner-side follow-up that contract's changeset explicitly deferred. It adapts to that contract rather than the earlier step-less-claim design:

- The claim still returns the job payload **plus** a `lease_token` at `POST /runners/jobs/request`; the runner reads the token and ignores `steps[]` (it pulls steps one at a time).
- Reports echo the dispatched `attempt` (so the workflow can reject stale reports) and carry the process `exit_code`.
- The workflow finalizes the job server-side (`WORKFLOWS_JOB_COMPLETED`), so the runner no longer calls `/complete`.

Runner-side hardening:

- A dedicated lease-token `ky` client (separate from the runner-token client used for claim/heartbeat), with idempotent POST retry — both `next` and `report` are idempotent, so a lost response is retried in place and a step is never re-pulled or re-executed.
- A step that crashes locally before producing a result (e.g. a temp-file write failure) is reported `failed`, so it never hangs `running`.
- The loop stops on `done` / `404` / `cancel` / abort.

Scope is runner-only (`apps/runner`); no `runners-dto`/route/changeset churn (the backend contract landed separately). New tests cover the loop and the two auth contexts; `test/env.ts` provides `SHIPFOX_API_URL` for the suite that exercises the real config.

Review focus: the idempotency/retry in `api-client.ts` and the abort/crash handling in `step-loop.ts`.

Refs ENG-399


ThierryAbalea:
The runner now stops when `nextStep` returns `{kind: 'done'}` and no longer calls `/runners/jobs/:jobId/complete`, but the workflow route descriptions still say "the runner must still call the job-completion endpoint". Could we update those descriptions for the interim ENG-399/ENG-400 protocol, or explicitly say finalization is currently handled by the follow-up server-side work?

ThierryAbalea (dismissed):
Reviewed.

PR #165, opened 2026-06-11
[api/integrations] Add Sentry install/connect flow

noe-charmet:
Adds the Sentry **install/connect flow** so a workspace member can link a Sentry installation to a Shipfox workspace. This is **PR2 of 2** — PR1 (#147) shipped the webhook receiver; without a way to create the installation + connection rows, every inbound webhook resolved to "unknown installation" and was dropped, so PR1 is inert in production until this lands.

## Why

PR1 normalizes Sentry issue webhooks and publishes `integrations.event.received`, but the receiver looks up an `integrations_sentry_installations` row that nothing created yet. PR2 adds the two routes that write that row + the core `integrations_connections` row, so resolving/unresolving a Sentry issue fans out to matching workflow trigger subscriptions.

## What

- **`@shipfox/api-integration-sentry`:** a stateless `SentryApiClient` (authorization-code exchange, org-slug derivation, optional verify-install), `handleSentryConnect`, and two authenticated routes — `POST /integrations/sentry/install` (returns the external-install URL) and `POST /integrations/sentry/connect` (links the installation). Provider HTTP failures map to a typed `SentryIntegrationProviderError`.
- **`@shipfox/api-integration-sentry-dto`:** install/connect request/response schemas.
- **`@shipfox/api-integration-core`:** wires the `getExistingSentryConnection` + `connectSentryInstallation` closures into the provider (internal wiring, no public API change).

## Decisions worth a careful review

- **No state param → POST /connect, session-authorized.** Sentry's install redirect carries no `state`, so the workspace comes from the request body and is authorized with `requireMembership`. `AUTH_USER` is **bearer-token** auth (not cookie), so authenticated POSTs are inherently CSRF-safe — no API-level CSRF guard. The residual lure risk (a member submitting an attacker's org tuple) is mitigated by the client confirmation screen in the follow-up UI ticket.
- **`org_slug` is derived from Sentry, not trusted from the body.** The connect body is `{workspace_id, code, installation_id}` only; the org slug comes from a `getInstallation` call after the exchange.
- **Persist before verify.** `verifyInstallation` (the `PUT status=installed` side effect) runs **after** the DB transaction, so a verify failure leaves a working connection rather than a Sentry-side "installed" state pointing at a row that was never written.
- **No secrets stored or logged.** No Sentry token is persisted; the error mapper never carries the token, OAuth code, or client secret into the thrown error's `cause`/`data` (asserted by a test).
- **Idempotent reconnect + disabled re-activation.** An already-active same-workspace connection short-circuits before the (single-use) code exchange; a previously-uninstalled (disabled) connection re-activates both the connection lifecycle and the installation status.

## Known limitation (deferred)

A concurrent same-install connect to two workspaces can repoint the installation row and orphan the loser's active connection. GitHub's provider has the identical pattern; the shared core-layer fix is tracked in ENG-409. Low probability (one install per org, unguessable id, both sides need a real installer).

## Out of scope (follow-up ticket)

Client UI (connect button, Sentry callback page, workspace-confirmation screen) and `/__e2e/integration-sentry` helper routes.

Refs ENG-409

---
## Summary by cubic
Add Sentry install/connect flow with `POST /integrations/sentry/install` and `POST /integrations/sentry/connect` to link a Sentry installation to a workspace and activate webhook handling.

- **New Features**
  - Introduced stateless `SentryApiClient` (code exchange, org-slug derivation, optional verify-install) with typed `SentryIntegrationProviderError`; tokens are never stored and secrets never leak in errors.
  - Added authenticated routes: `POST /integrations/sentry/install` (returns external install URL) and `POST /integrations/sentry/connect` (creates connection) with schemas in `@shipfox/api-integration-sentry-dto`; enforce `AUTH_USER` and workspace membership via `@shipfox/api-workspaces`.
  - Derive `org_slug` from Sentry; request accepts `{workspace_id, code, installation_id}` only. Persist connection + installation in one transaction before verify; verify-install is best-effort and non-fatal after persistence.
  - Idempotent reconnect and disabled re-activation; 409 when installation is linked to another workspace; provider errors map to 422/429/503.
  - Wired `getExistingSentryConnection` and `connectSentryInstallation` into `@shipfox/api-integration-core`. Concurrent connect race is tracked in ENG-409.

- **Dependencies**
  - Added `ky`, `@shipfox/api-auth-context`, and `@shipfox/api-workspaces`.

<sup>Written for commit 12a26124f190cfc2b3b4cf4838bb092f03927303. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/165?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libs/api/integration/sentry/src/core/install.ts
+++ b/libs/api/integration/sentry/src/core/install.ts
@@ -1,0 +1,75 @@
+import type {IntegrationConnection} from '@shipfox/api-integration-core-dto';
+import type {SentryApiClient} from '#api/client.js';
+import {SentryInstallationAlreadyLinkedError} from '#core/errors.js';
+
+export interface ConnectSentryInstallationInput {
+  workspaceId: string;
+  installationUuid: string;
+  orgSlug: string;
+  displayName: string;
+  installerUserId: string;
+}
+
+export interface HandleSentryConnectParams {
+  sentry: SentryApiClient;
+  workspaceId: string;
+  code: string;
+  installationUuid: string;
+  installerUserId: string;
+  verifyInstall: boolean;
+  getExistingSentryConnection: (input: {
+    installationUuid: string;
+  }) => Promise<IntegrationConnection<'sentry'> | undefined>;
+  connectSentryInstallation: (
+    input: ConnectSentryInstallationInput,
+  ) => Promise<IntegrationConnection<'sentry'>>;
+}
+
+// The "verify install" side effect runs AFTER the row is durably persisted, so a
+// verify failure leaves a working connection rather than a Sentry-side "installed"
+// state pointing at a row that was never written.
+export async function handleSentryConnect(
+  params: HandleSentryConnectParams,
+): Promise<IntegrationConnection<'sentry'>> {
+  const existing = await params.getExistingSentryConnection({
+    installationUuid: params.installationUuid,
+  });
+  if (existing && existing.workspaceId !== params.workspaceId) {
+    throw new SentryInstallationAlreadyLinkedError(params.installationUuid);
+  }
+  // Short-circuit so a replayed callback does not re-exchange the single-use code.
+  if (existing && existing.lifecycleStatus === 'active') {
+    return existing;
+  }
+
+  // Exchanging the code is also the authenticity check — a forged installation
+  // id fails here and nothing is persisted.
+  const authorization = await params.sentry.exchangeAuthorizationCode({
+    installationUuid: params.installationUuid,
+    code: params.code,
+  });
+
+  const {orgSlug} = await params.sentry.getInstallation({
+    installationUuid: params.installationUuid,
+    token: authorization.token,
+  });
+
+  // For a previously uninstalled (disabled) connection, this re-activates both the
+  // connection lifecycle and the installation status.
+  const connection = await params.connectSentryInstallation({
+    workspaceId: params.workspaceId,
+    installationUuid: params.installationUuid,
+    orgSlug,
+    displayName: `Sentry ${orgSlug}`,
+    installerUserId: params.installerUserId,
+  });
+
+  if (params.verifyInstall) {
+    await params.sentry.verifyInstallation({
+      installationUuid: params.installationUuid,
+      token: authorization.token,
+    });
+  }
+
+  return connection;
+}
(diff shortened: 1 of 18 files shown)


libs/api/integration/sentry/src/core/install.ts
```diff
@@ -0,0 +1,75 @@
...
+  // connection lifecycle and the installation status.
+  const connection = await params.connectSentryInstallation({
+    workspaceId: params.workspaceId,
+    installationUuid: params.installationUuid,
+    orgSlug,
+    displayName: `Sentry ${orgSlug}`,
+    installerUserId: params.installerUserId,
+  });
+
+  if (params.verifyInstall) {
```
ThierryAbalea, line 68: One verify-install recovery edge: `verifyInstallation` runs after the DB transaction, so if the Sentry `PUT status=installed` fails, the route returns an error but the connection is already active. A retry then hits the existing-active short-circuit above and returns the row without retrying verify, leaving Sentry potentially never marked installed. Should verify failure be logged/non-fatal after persistence, or should we persist a state that lets a retry attempt verification instead of short-circuiting?

ThierryAbalea (dismissed):
Reviewed.

libs/api/integration/sentry/src/core/install.ts
noe-charmet: Good catch — the persist-first ordering was supposed to make verify non-fatal, but I'd left `verifyInstallation` throwing, so it still errored the connect and a retry hit the active short-circuit without re-verifying.

Fixed in 12a2612: verify is now best-effort. After the row is persisted we attempt the verify and, on failure, log a structured warning (`sentry connect: verify-install failed after persistence, connection is active`) and return the active connection. The integration works regardless; only Sentry's UI-side "installed" flag may stay pending.

I went with log/non-fatal rather than the persist-a-state-and-re-verify-on-retry option because re-verifying without re-exchanging the (single-use) code needs a freshly minted installation token via the JWT-bearer grant, which is deliberately out of scope for this PR (we store no tokens yet). When that token-minting lands, a reconcile path/cron can clear any pending installs. Added a unit test covering the non-fatal path.

noe-charmet: Captured the durable-reliability follow-up as ENG-411 (https://linear.app/shipfox/issue/ENG-411). Recommended path: build `mintInstallationToken` (Sentry JWT-bearer grant, no stored secret) + a reconcile cron on the existing maintenance queue that re-verifies pending installs; an outbox→Temporal-workflow variant is written up there as the alternative if staging shows a pending install blocks webhook delivery.

PR #179, opened 2026-06-15
[client/integrations] Split integration gallery into Installed and Available sections

noe-charmet:
Redesign the integration gallery into two clearly separated sections on both the onboarding gallery page and the workspace settings hub: **Installed** (one card per connection) and **Available** (every installable provider).

The previous `IntegrationGallerySection` merged providers and connections into a single list and collapsed connections by provider via a `Map` keyed on `connection.provider`. That threw away every connection past the first for a given provider, even though the data model allows multiple connections per provider (e.g. two GitHub orgs) — a UI data-loss bug. A single `IntegrationGallery` now self-fetches both queries and renders the two sections directly; both pages render `<IntegrationGallery />`.

Key behavior:

- **Installed:** one card per connection (no collapse), stably sorted by provider display name then `created_at`, with the lifecycle badge, an "Added" subtitle, and an "Open in provider" link whose accessible label carries the account name. Disabled connections are dimmed.
- **Resilience:** installed cards never depend on the client catalog or the providers query — missing catalog entries fall back to a generic icon, and the provider label falls back to the raw provider key, so a providers-query failure renders correct labels instead of hiding the section.
- **Available:** every installable provider with a `Connect` link (re-install allowed), including already-connected ones. Providers with no catalog entry are skipped.
- **Onboarding suppression:** the Installed section (skeleton, error, and empty state included) is hidden until at least one connection loads. Settings always shows it, including the empty state.
- **Capability filter:** connections are filtered by capability in memory rather than via the hook, to avoid colliding with the active-only `source_control` query key in the React Query cache.

Error coordination is per-section: a providers failure blocks only Available, a connections failure blocks only Installed.

Refs ENG-437

---
## Summary by cubic
Split the integration gallery into Installed and Available with a denser, unified UI: Installed is a scannable list with right-aligned status pills; Available is a compact, responsive tile grid. Fixes the bug that hid extra connections per provider and aligns onboarding with ENG-437.

- **New Features**
  - Added `IntegrationGallery` that self-fetches and renders both sections.
  - Installed: one row per connection, stably sorted by provider then created_at; lifecycle badge with the status pill always rightmost; disabled rows are muted; visible action is “Open” (aria-label includes account and provider); meta shows “Added” date only; falls back to a generic icon/provider key; flat flex row.
  - Available: full-tile “Connect” link (no nested button), includes already-connected providers, skips providers missing a catalog entry, grid adapts to container and packs 3+ tiles where space allows; “Connect” affordance is muted and turns brand color on hover.
  - Onboarding: hide Installed (skeleton, error, and empty state) until at least one connection loads; settings always shows it, including the empty state; capability filter done in memory.

- **Bug Fixes**
  - Fixed data-loss bug that collapsed multiple connections for the same provider.
  - Errors are isolated per section; providers-query failures no longer hide Installed.

<sup>Written for commit 43339beaa0b634285893b7b13503376d101da9e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/179?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


noe-charmet:
image

PR #187, opened 2026-06-15
[runner] Shallow-clone the repository in the setup step

noe-charmet:
Performs the real repository checkout inside the runner's synthetic "Set up job" step. Until now the setup step only prepared the per-job workspace directory; this fills in the checkout body: ensure `git` is available, exchange the job lease for short-lived read-only checkout credentials via the checkout-token endpoint (ENG-406), and shallow-clone the project's default branch into the per-job directory.

This is the last piece that makes runner-dispatched jobs repository-aware. The "fail before steps" reporting channel already exists (the synthetic setup step from ENG-412), so a checkout failure now flows through the normal step protocol and fails the job in seconds instead of hanging until lease expiry.

## Failure handling

Every checkout failure mode fails the job before any user step runs, with a machine-readable `reason` persisted on the step:

- `git_unavailable` - git is not on the runner host's PATH (checked **before** minting a credential, so a host without git never hits the provider).
- `checkout_auth_failed` - credential exchange denied (401/403, or a provider `access-denied`/`forbidden`), or git rejected the credential.
- `checkout_unavailable` - provider unavailable: the backend's 429/503 retryable signal, or a git-side network failure (unreachable host / timeout). Mapped here for troubleshooting honesty; the retry policy stays a separate concern.
- `checkout_failed` - generic clone failure, missing checkout intent, and other token-endpoint errors.

## Credential handling

The credential is injected with a one-shot `git -c http.extraHeader`, never embedded in the clone URL, so it is **not** persisted to `.git/config` where later user steps running in the same cwd could read it (asserted by a real-git test). `GIT_TERMINAL_PROMPT=0` turns a denied credential into an immediate error instead of a hang. Token material is redacted from any error message. The shared redaction helper and credential-lifetime guarantees remain owned by ENG-408 (which this unblocks); this PR ships a local redactor.

## Shape

Per an engineering-review pass, the whole setup phase stays cohesive in `executeSetupStep` (token fetch + git + clone + reason mapping). `execution` gains a dependency on `protocol`, which does not violate the ENG-410 rule (nobody imports orchestration; only protocol touches the wire) and naturally orders the git check ahead of credential minting.

## Review notes

- `workspace/checkout.ts` is the new module (git invocation, stderr classification, redaction).
- Tests use a real local `file://` git remote for the happy path, credential non-persistence, and bad-ref failure; `execFile` is mocked only for the network/auth/abort paths a local remote cannot produce.
- One known edge, deliberately deferred: an unscoped `extraHeader` would ride a cross-origin redirect. GitHub/debug providers do not redirect cross-host, so it is low risk for MVP; per-host scoping is a natural ENG-408 follow-up.

Closes ENG-405

---
## Summary by cubic
Shallow-clone the repository during the runner’s “Set up job” step using short‑lived checkout tokens, with fast, machine‑readable failure reasons. Implements ENG‑405 to make jobs repository‑aware and prevent hangs before steps run.

- **New Features**
  - Check for `git` before minting credentials; fail with `git_unavailable`.
  - Exchange the lease for checkout credentials via `requestCheckoutToken`, then shallow‑clone the default branch into the per‑job dir.
  - Map failures to stable reasons: `checkout_auth_failed`, `checkout_unavailable`, `checkout_failed`, and `setup_aborted`; workspace prep errors use `workspace_prep_failed`.
  - Inject credentials with a one‑shot `git -c http.extraHeader`; never persist to `.git/config`; redact tokens in errors.
  - Abort‑aware: kill the git child via the job signal and report `setup_aborted` when applicable.

- **Bug Fixes**
  - Correctly classify checkout‑token errors using `ky`’s `HTTPError.data`: 422 `access‑denied`/`forbidden` → `checkout_auth_failed`; 429/503 and provider codes (`rate‑limited`, `timeout`, `provider‑unavailable`) → `checkout_unavailable`.
  - Harden redaction: remove all wire forms of credentials (including base64 for basic auth) and scrub the original exec error from the cause chain.
  - Widen git stderr matching to cover real 401/403, 5xx, and 429 rate‑limit messages.

<sup>Written for commit 2205c4574b09e4380bfb77a84df95ca1c02df75a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libs/runner/workspace/src/checkout.ts
+++ b/libs/runner/workspace/src/checkout.ts
@@ -1,0 +1,131 @@
+import {execFile} from 'node:child_process';
+import {promisify} from 'node:util';
+import type {CheckoutTokenAuthDto} from '@shipfox/api-workflows-dto';
+
+const execFileAsync = promisify(execFile);
+
+/** Thrown when `git` is not on the runner host's PATH; surfaced as `git_unavailable`. */
+export class GitUnavailableError extends Error {
+  constructor(options?: ErrorOptions) {
+    super('git is not available on the runner host', options);
+    this.name = 'GitUnavailableError';
+  }
+}
+
+/**
+ * Why a clone failed, kept abstract here so the workspace layer does not depend on the
+ * step-error DTO. The setup step maps each kind to a machine-readable `reason`.
+ */
+export type CheckoutFailureKind = 'auth' | 'unavailable' | 'failed' | 'aborted';
+
+export class CheckoutError extends Error {
+  constructor(
+    public readonly kind: CheckoutFailureKind,
+    message: string,
+    options?: ErrorOptions,
+  ) {
+    super(message, options);
+    this.name = 'CheckoutError';
+  }
+}
+
+/** Throws {@link GitUnavailableError} when `git` cannot be invoked on the host. */
+export async function assertGitAvailable(): Promise<void> {
+  try {
+    await execFileAsync('git', ['--version']);
+  } catch (error) {
+    throw new GitUnavailableError({cause: error});
+  }
+}
+
+/**
+ * Removes every occurrence of each secret plus any URL-embedded `user:pass@` credential
+ * from `text`, so a clone error never carries token material into a log or step result.
+ */
+export function redactSecrets(text: string, secrets: string[]): string {
+  let redacted = text;
+  for (const secret of secrets) {
+    if (secret) redacted = redacted.split(secret).join('***');
+  }
+  return redacted.replace(/(https?:\/\/)[^/@\s]+@/gi, '$1***@');
+}
+
+// git surfaces credential rejection and provider-availability failures only through
+// stderr text, so classification is pattern-based. Auth wins over unavailable when both
+// could match, since a rejected credential is the more actionable cause.
+const AUTH_FAILURE =
+  /authentication failed|could not read username|invalid username or password|terminal prompts disabled|403 forbidden|permission denied \(publickey\)|access denied/i;
+const PROVIDER_UNAVAILABLE =
+  /could not resolve host|could not connect|connection timed out|failed to connect|temporary failure in name resolution/i;
+
+/**
+ * Shallow-clones `ref` of `repositoryUrl` into `cwd` (which must already exist and be empty).
+ *
+ * The credential is injected with a one-shot `-c http.extraHeader`, never embedded in the
+ * clone URL, so it is not persisted to `.git/config` where later user steps could read it.
+ * `GIT_TERMINAL_PROMPT=0` turns a missing or denied credential into an immediate error
+ * instead of a hang on an interactive prompt. Failures are classified into a
+ * {@link CheckoutError} and have any token material redacted from their message.
+ */
+export async function checkoutRepository(params: {
+  repositoryUrl: string;
+  ref: string;
+  auth?: CheckoutTokenAuthDto | undefined;
+  cwd: string;
+  signal?: AbortSignal | undefined;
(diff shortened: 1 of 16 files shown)


libs/runner/workspace/src/checkout.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+  return redacted.replace(/(https?:\/\/)[^/@\s]+@/gi, '$1***@');
+}
+
+// git surfaces credential rejection and provider-availability failures only through
+// stderr text, so classification is pattern-based. Auth wins over unavailable when both
+// could match, since a rejected credential is the more actionable cause.
+const AUTH_FAILURE =
+  /authentication failed|could not read username|invalid username or password|terminal prompts disabled|403 forbidden|permission denied \(publickey\)|access denied/i;
+const PROVIDER_UNAVAILABLE =
```
ThierryAbalea, line 58: Git often reports HTTP failures as `The requested URL returned error: 403/429/503`; those strings do not match either classifier today, so denied credentials or provider outages can fall through to `checkout_failed`. Could we add status-based patterns/tests so 401/403 maps to `checkout_auth_failed` and retryable 429/5xx maps to `checkout_unavailable`?

noe-charmet: Good catch, fixed. `classifyCloneError` now matches git's `The requested URL returned error: <status>` form in addition to the prose messages:

- `40[13]` (401/403) → `AUTH_FAILURE` → `checkout_auth_failed`
- `429` and `5\d\d` → `PROVIDER_UNAVAILABLE` → `checkout_unavailable`

with tests for the git-side 403, 429, and 503 cases. This aligns the stderr-based classification with the token-endpoint path (`classifyCheckoutTokenError`), which already maps 401/403 → auth and 429/503 → unavailable by HTTP status.

- 401/403 + 5xx: 29b7301c
- 429: 2205c457

libs/runner/workspace/src/checkout.ts
```diff
@@ -0,0 +1,131 @@
...
+  ref: string;
+  auth?: CheckoutTokenAuthDto | undefined;
+  cwd: string;
+  signal?: AbortSignal | undefined;
+}): Promise<void> {
+  const {repositoryUrl, ref, auth, cwd, signal} = params;
+
+  const args: string[] = [];
+  if (auth) {
+    args.push('-c', `http.extraHeader=Authorization: ${authorizationValue(auth)}`);
```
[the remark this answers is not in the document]
noe-charmet: Acknowledged. This is a deliberate deferral, not an oversight (see the PR description's "Review notes").

Per-host scoping (binding the header to the repository origin, e.g. `http.https://github.com/.extraHeader`, like `actions/checkout`, so a cross-origin redirect drops it) is tracked as the ENG-408 follow-up. For this MVP the residual risk is low: the providers in play (GitHub, debug source control) do not redirect cross-host, the credential is short-lived and read-only, and it is injected one-shot via `-c` (never persisted to `.git/config`). We'll scope the header in ENG-408.

ThierryAbalea (approved):
Reviewed current head; previous checkout-classification concern is fixed and CI is green.

PR #193, opened 2026-06-15
Add workflow run graph inspector

ThierryAbalea:
Adds the Phase 5 run detail graph and inspector behavior on top of the Phase 4 dashboard frame.

The frame previously selected the failed job implicitly and rendered placeholder step content. This makes job, step, and attempt selection interactive so a failed workflow run can be triaged from the overview before opening logs.

The implementation intentionally keeps Logs as a placeholder and uses the existing run-detail DTO blobs for gate, output, restart, and error display. Cycle ribbons, restart edges, richer log/source panes, copy controls, and key-value gate rendering are left for the logs/source/fidelity phase.

Design comparison was performed against the extracted Claude Design React source and preview screenshot. Claude consult findings around keyboard focus and preserving the active inspector tab across selection changes were applied before commit.

## Stack

Part 1 of 2.

Depends on: none
Next: #194

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #193 | `workflow-execution-dashboard/05-graph-inspector` | `workflow-execution-dashboard/04-run-detail-frame` |
| 2 | #194 | `workflow-execution-dashboard/06-logs-source-fidelity` | `workflow-execution-dashboard/05-graph-inspector` |

PR #194, opened 2026-06-15
Add workflow logs and source fidelity

ThierryAbalea:
Adds the Phase 6 logs/source fidelity pass on top of the interactive run detail graph and inspector.

The dashboard now has a frontend-only Logs tab with fixture log rows, stream filtering, search, selected-attempt preservation, and an honest empty state for steps without attempts. This intentionally does not add a logs backend, API route, or DTO contract.

The Source tab now renders the run snapshot bundle: YAML by default, with secondary document/model views. It uses only `workflow_source_yaml`, `workflow_document`, and `workflow_model` from the run detail DTO, so old runs without a snapshot show an unavailable state rather than falling back to the mutable current definition.

Design comparison was performed against the extracted Claude Design React source and refreshed Overview/Logs/Source screenshots. Claude consult findings for root-cause focus behavior, removing fixture-status UI text, live log copy, and missing tests were applied before commit.

## Stack

Part 2 of 2.

Depends on: #193
Next: none

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 | #193 | `workflow-execution-dashboard/05-graph-inspector` | `workflow-execution-dashboard/04-run-detail-frame` |
| 2 (current) | #194 | `workflow-execution-dashboard/06-logs-source-fidelity` | `workflow-execution-dashboard/05-graph-inspector` |

### DIFF
--- a/libs/client/projects/src/pages/project-run-detail-page.test.tsx
+++ b/libs/client/projects/src/pages/project-run-detail-page.test.tsx
@@ -12,8 +12,11 @@ const BUILD_JOB_RE = /^build\s+succeeded/;
 const ARTIFACT_OUTPUT_RE = /"artifact": "dist\/app\.tar\.gz"/;
 const LOCKFILE_OUTPUT_RE = /"lockfile": "stale"/;
 const ATTEMPT_1_RE = /Attempt 1 failed/;
+const SELECT_ATTEMPT_1_RE = /Select attempt 1 failed/;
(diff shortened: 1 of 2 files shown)


ThierryAbalea:
Closing this previous workflow execution dashboard stack. We are restarting with a new implementation strategy that ports directly from the exported Claude Design JSX for higher fidelity.

PR #196, opened 2026-06-16
Add log ingest module foundation

noe-charmet:
Introduces a stateless `log-ingest` module in the API monolith — the foundation every other workstream of the runner log capture & streaming project builds on. It captures step log chunks from runners, accounts them against a per-job budget, and stores them hot in Postgres pending later compaction.

## What this adds

- **New module `@shipfox/api-log-ingest`** with its own schema: per-job budget accounting, per-attempt stream lifecycle, and hot chunk storage (`bytea`), plus the module outbox table.
- **Offset-CAS append endpoint** `POST /runs/jobs/current/steps/:stepId/logs?attempt=N&offset=B`, authenticated by the existing job lease token. Append succeeds only when `offset == committed_length`; earlier offsets are acked as already applied (idempotent retries), later offsets are rejected `409` with the committed offset so the runner rewinds. No step-state gate — the lease alone authorizes writes.
- **Per-job accrual budget** on decoded payload bytes (`5MB base + 1MB/min`, configurable, no hard ceiling). At cap the server injects a `capped` control tombstone, drops further output, and signals the runner via the response.
- **NDJSON v1 record contract** (output + 6 control kinds) in the sibling `@shipfox/api-log-ingest-dto` package.
- **S3-compatible client + config** targeting the bundled **Garage** (added to `compose.yml`, with `dev/garage/bootstrap.sh` and a committed `.env`). A readiness check gates on bucket reachability without crashing liveness.
- `@shipfox/node-fastify` gains a `createRawBodyPlugin({contentType, bodyLimit})` factory (the webhook raw-body plugin now builds on it) for byte-exact request bodies.

## Notable design choices

- **Streams are keyed `(job_id, step_id, attempt)` and scoped to the lease's job**, so a lease can only ever reach its own job's streams — cross-job log injection is structurally impossible, with no query into the workflows tables.
- **Lock-free**: the CAS and budget accrual are atomic conditional `UPDATE`s (no `FOR UPDATE`), so concurrent and multi-instance appends serialize through Postgres with no deadlock surface and no per-job serialization point. The body is parsed and validated before the transaction opens, so a malformed or large body never holds a row.
- `committed_length` counts only runner bytes; server-injected control records carry `kind='control'`, do not advance it, and order by `seq` for reads.

## Review notes

- The byte-axis split is the load-bearing invariant: `offset`/`committed_length` are positions in the raw NDJSON spool stream; the budget counts decoded `data` UTF-8 bytes only. See the diagram on `db/schema/chunks.ts`.
- **Operational**: hot chunks are bounded per job by the budget but not across jobs until compaction + retention (ENG-442) ship. Runner log capture should stay disabled in production until then, or Postgres grows unbounded.

This is the foundation that unblocks the lifecycle workers (ENG-442) and the read path + viewer (ENG-443).

Closes ENG-439

---
## Summary by cubic
Adds a stateless `@shipfox/api-logs` module with an offset‑CAS append endpoint, a per‑job stored‑byte budget, and hot chunk storage to found runner log capture and streaming (ENG‑439). Renames the module to logs, auto‑provisions local Garage, cascades chunk deletes, and documents bigint‑aggregate guidance.

- **New Features**
  - Append endpoint: `POST /runs/jobs/current/steps/:stepId/logs?attempt=N&offset=B` (lease‑token auth). Accepts only `offset == committed_length`; earlier offsets are acked; later or straddling appends return 409 with committed length. Empty bodies heartbeat (no stream creation). Stamps lease workspace/project/run on stream rows; lease/stream identity mismatches return 403 without partial updates.
  - Budget and contract: budget charges stored NDJSON bytes (framing + control included). At cap the server injects a `capped` control and drops further output. NDJSON v1 contract in `@shipfox/api-logs-dto` bounds record sizes, rejects empty `output`, and caps `group_start` name length.
  - Schema and Dev/Ops: FK from `logs_chunks.stream_id` to `logs_attempt_streams.id` with ON DELETE CASCADE. Compose bundles Garage and auto‑provisions it; S3 client defaults to local Garage with fail‑fast timeouts; API readiness no longer gates on bucket reachability; `createRawBodyPlugin({contentType, bodyLimit})` in `@shipfox/node-fastify`; fixed outbox index name collisions. Documented bigint‑aggregate rule: per‑row byte counts use number; cross‑row aggregates must read as bigint.

- **Migration**
  - Local: run `docker compose up -d` — Garage and its bootstrap run automatically; `.env` defaults to the local Garage.
  - Deploy: run DB migrations and configure your S3 bucket and real credentials. Keep runner log capture disabled in production until compaction/retention land.

<sup>Written for commit 81b82e3ef49f598e139abb180236ad26140c3a01. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/196?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libs/api/logs-dto/src/schemas/record.ts
```diff
@@ -0,0 +1,69 @@
...
+ *   - offset / committed_length  → byte positions in the RAW NDJSON spool stream
+ *     (envelope included). This is the offset-CAS axis the runner tracks.
+ *   - budget payload bytes        → decoded `data` UTF-8 bytes of `output` records
+ *     only (envelope and control records excluded). This is the accrual-budget axis.
+ */
+
+/** Largest decoded `data` payload per record. Longer lines are split by the runner. */
+export const MAX_RECORD_DATA_BYTES = 16 * 1024;
+
+const baseRecordShape = {
```
ThierryAbalea, line 22: #196 and #197 currently define the same log append/record wire contract in two packages (`@shipfox/api-log-ingest-dto` here vs `@shipfox/api-logs-dto` in the runner PR), and the schemas already differ (`src` optional/defaulting, payload-size validation, helpers/constants). Could we make one DTO package own the contract and have both ingest and runner import it before these land?

apps/api/src/core/run.ts
```diff
@@ -51,7 +53,13 @@ export async function run(): Promise<void> {
   const mountedE2eRoutes = createE2eRouteGroup(e2eRoutes, config);
 
   logger().info('Creating HTTP server');
-  await createApp({auth: [...auth, ...e2eAuth], routes: [...routes, ...mountedE2eRoutes]});
+  await createApp({
+    auth: [...auth, ...e2eAuth],
+    routes: [...routes, ...mountedE2eRoutes],
+    // Readiness (not liveness) gates on log object storage so a Garage outage
+    // marks the instance not-ready without crashing the process.
+    readinessChecks: [{name: 'log-storage', check: checkBucketReachable}],
```
ThierryAbalea: The new `/readyz` check depends on `checkBucketReachable`, but Garage’s bucket/key are created only by `dev/garage/bootstrap.sh`; the E2E startup path appears to start compose and then wait on `/readyz`, so CI times out before the bucket exists. Since Linear makes object storage a required dependency, could we wire the Garage bootstrap into the dev/E2E/CI setup so a fresh stack can become ready?

libs/api/logs/src/presentation/routes/append-logs.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const appendLogsRoute = defineRoute({
+  method: 'POST',
+  path: '/steps/:stepId/logs',
+  description:
+    'Appends a chunk of framed NDJSON log records for one step attempt of the leased job. The body is raw NDJSON bytes (whole records, newline-terminated). `offset` must equal the server-held committed length: an earlier offset is acknowledged as already applied, a later offset is rejected with the committed length so the runner rewinds. The job is identified by the lease token; there is no step-state gate. When the per-job log budget is exhausted the response reports `capped` and further output is dropped.',
+  schema: {
+    params: z.object({stepId: z.string().uuid()}),
+    querystring: appendLogsQuerySchema,
+    response: {
```
ThierryAbalea, line 20: The runner parses the 409 body to rewind its log offset, and `offsetGapResponseSchema` already exists, but the route only declares the 200 response. Could we add the 409 schema here too so this load-bearing conflict contract is documented and route-validated?

noe-charmet:
Thank for the early review, however it was not yet marked as ready for review, still had some passes to go through.

libs/api/logs-dto/src/schemas/record.ts
noe-charmet: Right, the duplication landed because #196 and #197 were authored in parallel. This PR has since been narrowed down to the single canonical package (renamed `@shipfox/api-logs-dto` in 54e4e61e, was `log-ingest-dto`). The runner only ships logs in a follow-up — when it does, it'll import from `@shipfox/api-logs-dto` rather than vendoring its own copy. Closing here as fixed on this side; the runner-side will pick it up when it lands.

apps/api/src/core/run.ts
noe-charmet: Good catch — addressed in 669901e8 ("Provision local Garage automatically on docker compose up"). The compose file now ships a `garage-init` service that runs `dev/garage/bootstrap.sh` and gates the `api` service on its healthy completion, so a fresh stack reaches `/readyz` without manual bootstrap.

Separately, e1b2f4af stopped the `/readyz` probe from gating on object-storage reachability, so even a transient Garage hiccup no longer pulls the API instance out of rotation — readiness only fails when something the request path itself can't tolerate is down.

libs/api/logs/src/presentation/routes/append-logs.ts
noe-charmet: Done in e1b2f4af. The route now declares both `200: appendLogsResponseSchema` and `409: offsetGapResponseSchema`, so the offset-gap contract is route-validated and shows up in the OpenAPI surface the runner consumes.

libs/api/logs-dto/src/schemas/record.ts
noe-charmet: The other PR will rebase on this one once merged and will remove duplicaiton. The other one is still under draft status for this exact reason as it has some implicit dependency here.

libs/api/logs/src/config.ts
```diff
@@ -0,0 +1,36 @@
...
+  }),
+  LOG_STORAGE_S3_SECRET_ACCESS_KEY: str({
+    desc: 'Secret access key used to authenticate to the object store. Defaults to the local-development Garage secret; set real credentials for production.',
+    default: '0000000000000000000000000000000000000000000000000000000000000000',
+  }),
+  LOG_STORAGE_S3_FORCE_PATH_STYLE: bool({
+    desc: 'Whether to address the bucket as a path (endpoint/bucket) instead of a subdomain. Set it to true for Garage and MinIO; false works for AWS S3.',
+    default: true,
+  }),
+  LOG_BUDGET_BASE_BYTES: num({
```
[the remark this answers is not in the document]
noe-charmet: Right, the defaults are 1024-based (5 × 1024² and 1 × 1024², so MiB). Fixed in 5bfdfb9e: descriptions now say "5 MiB" / "1 MiB per minute". Kept the values as-is since the power-of-two budget is conventional for byte budgets.

libs/api/logs/drizzle/0000_initial.sql
```diff
@@ -0,0 +1,50 @@
...
+	"declared_total_bytes" bigint,
+	"truncated" boolean DEFAULT false NOT NULL,
+	"object_key" text,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE "logs_chunks" (
+	"id" uuid PRIMARY KEY DEFAULT uuidv7() NOT NULL,
+	"stream_id" uuid NOT NULL,
```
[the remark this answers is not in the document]
noe-charmet: Good catch — fixed in 5bfdfb9e. Added `references(() => attemptStreams.id)` on `logs_chunks.stream_id`, matching the workflows-module convention (no cascade, so the DB refuses to delete a parent stream while children exist).

libs/api/logs/src/db/streams.ts
```diff
@@ -0,0 +1,128 @@
...
+export async function getOrCreateAttemptStream(
+  tx: Transaction,
+  identity: AttemptStreamIdentity,
+): Promise<AttemptStream> {
+  const [row] = await tx
+    .insert(attemptStreams)
+    .values(identity)
+    .onConflictDoUpdate({
+      target: [attemptStreams.jobId, attemptStreams.stepId, attemptStreams.attempt],
+      set: {updatedAt: sql`now()`},
```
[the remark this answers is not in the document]
noe-charmet: The order is "upsert then assert", but `getOrCreateAttemptStream` is only ever called inside the `db().transaction(...)` block in `appendLogs`, so throwing `LeaseStreamMismatchError` rolls back the entire transaction — including the `updated_at` write from the conflict update. So a rejected lease cannot tick the stream's freshness.

5bfdfb9e tightens the existing mismatch test to lock that in: it now captures the row's `updated_at` and `committed_length` before the rejected append and asserts they're untouched after. If anyone ever pulls `getOrCreateAttemptStream` out of a transactional caller in the future, this test will fail and force a rethink.

libs/api/logs/src/db/schema/job-accounting.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Accrual-budget state per job run, keyed by the job id carried in the lease.
+ * `stored_bytes_used` counts the raw NDJSON bytes the server has stored for the
+ * job (envelope and control records included), so the budget bounds exactly what
+ * lands in Postgres. `started_at` is the budget clock origin (first append).
+ * `capped_at`, once set, makes every later append a no-op drop.
+ */
+export const jobAccounting = pgTable('job_accounting', {
+  jobId: uuid('job_id').primaryKey(),
+  workspaceId: uuid('workspace_id').notNull(),
+  storedBytesUsed: bigint('stored_bytes_used', {mode: 'number'}).notNull().default(0),
```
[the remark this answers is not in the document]
noe-charmet: JS `number` is safe (IEEE-754 integer-safe up to 2^53 ≈ 9 PB) because the per-job byte budget is bounded several orders of magnitude below that ceiling: the default `LOG_BUDGET_BASE_BYTES` is 5 MiB plus 1 MiB/min, capped enforcement stops growth after the budget crossing, and a single append body is hard-limited to 1 MiB by the route. `stored_bytes_used` (and the matching `committed_length` and `declared_total_bytes` on `attempt_streams`) therefore cannot reach the 2^53 precision boundary in any realistic job.

`mode: 'number'` is used deliberately across this module for that reason — keeps the domain layer working with plain numbers instead of bigint, and the dto schema even calls this out explicitly (see the comment on `offset` / `committed_length` in libs/api/logs-dto). Switching to `mode: 'bigint'` would force bigint arithmetic through every consumer for no benefit at these sizes; if budgets ever cross into PB territory we'll have bigger problems than precision.

libs/api/logs/drizzle/0000_initial.sql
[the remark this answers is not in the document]
noe-charmet: Follow-up in 83bd008b: switched the FK to `ON DELETE cascade`. `logs_chunks` is the spool for its parent stream with no independent lifecycle, so a stream purge (retention, project deletion, abandoned-job cleanup) should take its chunks with it — single statement instead of "delete chunks, then delete stream" boilerplate at every call site. Closer analog is the `@shipfox/api-auth` user-owned tables (`refresh_tokens`, `email_verifications`, `password_resets`), which all cascade.

libs/api/logs/src/db/schema/job-accounting.ts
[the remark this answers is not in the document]
noe-charmet: Follow-up: my earlier reply was partial — bounded *per row*, but a system-wide aggregate (e.g. "total bytes of logs stored") has no such bound and could plausibly cross 2^53 (~9 PB) at scale.

In Drizzle the column's `mode: 'number'` only controls *per-row* reads; an aggregate like `SUM(...)` is a separate SQL expression returning `NUMERIC`, with its own client-side coercion controlled at the query site. So the column declaration and the aggregate read are independent decisions:

```ts
sql<bigint>`SUM(${jobAccounting.storedBytesUsed})`  // safe to 2^63 regardless of column mode
```

Keeping the column as `number` lets the hot path (`accrued.used <= allowed` in `appendLogs`) stay plain-number arithmetic where the value is genuinely bounded by the per-job budget; aggregates that need the global sum opt into bigint at the query site, which is where it belongs.

Landed in 99f37eff: added the convention as a schema-level comment on `job_accounting.stored_bytes_used` and `attempt_streams.{committed_length,declared_total_bytes}` so a future aggregate writer sees it where the column lives, not buried in a PR thread.

noe-charmet:
Will be merging this one in order to progress on the next elements.

PR #199, opened 2026-06-16
Add workflow run detail contract

ThierryAbalea:
Adds the shared workflow run-detail contract needed by the execution dashboard before any UI work lands.

This moves the `GET /workflows/runs/:id` response shape into `@shipfox/api-workflows-dto`, includes nested jobs, steps, and attempts, and adds explicit `duration_ms` fields on runs, jobs, steps, and attempts. The backend returns `0` for duration values until real elapsed-duration computation is implemented; later UI adapters must derive pending/cancelled display state from status rather than treating `0` as unknown.

Also adds the client transport/hook and a small duration formatter used by the upcoming JSX-port phases.

Reviewed with Claude consult before commit: `/Users/thierry.abalea/dev/claude-consult-kit/consults/workflow-dashboard-jsx-port-phase-1-contract-review/agent-reviewer.md`.

## Stack

Part 1 of 2.

Depends on: none
Next: #200

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #199 | `workflow-dashboard-jsx-port/01-run-detail-contract` | `main` |
| 2 | #200 | `workflow-dashboard-jsx-port/02-source-snapshot-contract` | `workflow-dashboard-jsx-port/01-run-detail-contract` |

---
## Summary by cubic
Adds a run detail API contract and exposes durations. GET `/workflows/runs/:id` now returns jobs with steps and attempts, and all include `duration_ms`.

- **New Features**
  - DTOs: added `RunDetailResponseDto`/`runDetailResponseSchema` with jobs → steps → attempts.
  - Durations: added `duration_ms` to `RunDto`, `JobDto`, `StepDto`, and `StepAttemptDto` (currently `0`).
  - API: `/workflows/runs/:id` returns the new nested payload; run lists include `duration_ms`.
  - Client: added `getWorkflowRun` and `useWorkflowRunQuery` with `workflowRunsQueryKeys.detail`.
  - Utils: added `humanDurationMs(duration_ms)` for display.

- **Refactors**
  - `get-run` route now uses shared `runDetailResponseSchema` from `@shipfox/api-workflows-dto` instead of an inline schema.

<sup>Written for commit 9299f5d8cca51a85cd596eda40950584db76d0e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/199?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libs/api/workflows/src/presentation/dto/job.ts
```diff
@@ -9,6 +9,7 @@ export function toJobDto(job: Job): JobDto {
     status: job.status,
     dependencies: job.dependencies,
     position: job.position,
+    duration_ms: 0,
```
noe-charmet, line 12: This is de-facto making all workflows have a `0ms` value.
If this is a needed data for us, should we not implement a complete capture and storage of this data instead of returning a placeholder ?

libs/client/projects/src/lib/human-duration.ts
```diff
@@ -23,6 +23,20 @@ export function humanDuration(fromIso: string, toIso?: string): string {
   return `${totalHr}h ${pad2(min)}m`;
 }
 
+export function humanDurationMs(durationMs: number): string {
```
noe-charmet, line 26: Its kind of weird to have this unused utility function in this PR with other unrelated work in some way.

I would also suggest extracting these time manipulation functions for UI purpose to utilities available in our shared UI library. This would enforce stronger sharing. I see we also have a `formatDate` function that appears multiple time in the codebase and could be extracted together in a single change.

noe-charmet, line 26: As we use [date-fns](https://date-fns.org/v4.4.0/docs/intervalToDuration) quite heavily for date manipulation. we could do something like this, which would accept any unit of duration as input and display things correctly on any kind of durations (years, months, days, hours, minutes, seconds, milliseconds)
```typescript
import {
  add,
  formatDuration,
  intervalToDuration,
  type Duration,
} from "date-fns";

export function formatNormalizedDuration(
  duration: Duration,
  options?: Parameters<typeof formatDuration>[1],
): string {
  const start = new Date(0);

  const normalized = intervalToDuration({
    start,
    end: add(start, duration),
  });

  return formatDuration(normalized, options);
}
```

ThierryAbalea:

PR #200, opened 2026-06-16
Snapshot workflow source for runs

ThierryAbalea:
Adds immutable workflow source snapshots for execution dashboard source inspection.

Definitions now carry `workflow_source_yaml` alongside `workflow_document` and `workflow_model`, and run creation snapshots the source YAML, parsed document, and normalized model onto `workflow_runs`. Run detail returns only that run snapshot bundle, preserving legacy rows as `null` rather than reading mutable current definition state.

This keeps the upcoming JSX-port `SourceTab` backed by the exact source that produced the run. Line-range highlighting (`yamlLines` in the design export) remains an explicit adapter/UI follow-up.

Reviewed with Claude consult before commit: `/Users/thierry.abalea/dev/claude-consult-kit/consults/workflow-dashboard-jsx-port-phase-2-source-snapshot-review/agent-reviewer.md`. Applied the snapshot-construction finding so `runWorkflow` assembles the full immutable snapshot and the DB layer stores it verbatim.

Note: the affected test gate initially exposed a `@shipfox/runner-execution` SIGKILL timing flake under parallel Turbo execution. The package passed in isolation, and the same affected gate passed with `--concurrency=1`.

## Stack

Part 1 of 2.

Depends on: none
Next: #201

| Part | PR | Branch | Base |
| --- | --- | --- | --- |
| 1 (current) | #200 | `workflow-dashboard-jsx-port/02-source-snapshot-contract` | `workflow-dashboard-jsx-port/01-run-detail-contract` |
| 2 | #201 | `workflow-dashboard-jsx-port/03-design-component-port` | `workflow-dashboard-jsx-port/02-source-snapshot-contract` |

### DIFF
--- a/libs/api/definitions/src/core/entities/workflow-definition.ts
+++ b/libs/api/definitions/src/core/entities/workflow-definition.ts
@@ -4,6 +4,7 @@ import type {WorkflowModel} from './workflow-model.js';
 export type WorkflowSpec = WorkflowDocument;
 
 export interface WorkflowDefinitionPayload {
+  sourceYaml?: string | null;
   document: WorkflowDocument;
   model: WorkflowModel;
 }
@@ -21,6 +22,7 @@ export interface WorkflowDefinition {
    * before they migrate to `document`/`model`.
    */
   definition: WorkflowSpec;
+  sourceYaml: string | null;
   document: WorkflowDocument;
   model: WorkflowModel;
   contentHash: string | null;
--- a/libs/api/workflows/drizzle/0002_definition_snapshot.sql
+++ b/libs/api/workflows/drizzle/0002_definition_snapshot.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "workflows_workflow_runs" ADD COLUMN "definition_snapshot" jsonb;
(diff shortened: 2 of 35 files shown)


libs/api/workflows/drizzle/0002_definition_snapshot.sql
noe-charmet, line 1: You can edit the source migration directly for now, we have not deployed anything yet so we drop/recreate everything.

libs/api/definitions/src/core/entities/workflow-definition.ts
```diff
@@ -4,6 +4,7 @@ import type {WorkflowModel} from './workflow-model.js';
 export type WorkflowSpec = WorkflowDocument;
 
 export interface WorkflowDefinitionPayload {
+  sourceYaml?: string | null;
```
noe-charmet, line 7: Could we make things slightly forward thinking and avoid mentioning YAML explicitely if we ever want to support other source definition format.
Or at least have something like `{source: string, format: string}` ?

noe-charmet:
BTW this branch can most likely target main directly. Here the two changes in the stack seem to be pretty unrelated and separately scoped.

ThierryAbalea:

PR #201, opened 2026-06-16
Add workflow run dashboard

ThierryAbalea:
Add the workflow run dashboard surface for monitoring workflow executions.

The dashboard gives reviewers and operators a single view of recent runs, the selected run status, trigger metadata, duration, job dependencies, step execution state, restart context, logs, and workflow source. It uses fixture-backed workflow execution data so the frontend can exercise the complete monitoring experience before live route data is connected.

The page is split into reusable dashboard components and uses the shared Shipfox React UI primitives and Tailwind token classes throughout, keeping styling consistent with the rest of the client.

### DIFF
--- a/libs/client/projects/src/pages/workflow-dashboard/components/dashboard-shell.tsx
+++ b/libs/client/projects/src/pages/workflow-dashboard/components/dashboard-shell.tsx
@@ -1,0 +1,62 @@
+import {Avatar, Button, Code, Icon, Kbd, Logo, Text} from '@shipfox/react-ui';
+import type {ReactNode} from 'react';
+import {WorkflowDashboardProjectTabs} from './project-tabs.js';
+
+export function WorkflowDashboardShell({
+  children,
+  footerAction,
+}: {
+  children: ReactNode;
+  footerAction?: ReactNode;
+}) {
+  return (
+    <div className="grid h-screen grid-rows-[56px_40px_1fr] overflow-hidden bg-background-neutral-background text-foreground-neutral-base">
+      <WorkflowDashboardTopNav />
+      <WorkflowDashboardProjectTabs />
+      <div className="min-h-0 overflow-hidden">{children}</div>
+      {footerAction}
+    </div>
+  );
+}
+
+function WorkflowDashboardTopNav() {
+  return (
+    <header className="flex items-center justify-between gap-12 border-border-neutral-base border-b bg-background-components-base px-16">
+      <div className="flex min-w-0 items-center gap-8">
+        <Logo className="h-18 w-auto shrink-0" variant="wordmark" />
+        <span className="text-foreground-neutral-disabled text-lg">/</span>
+        <button
+          className="flex h-28 items-center gap-4 rounded-6 px-6 text-foreground-neutral-base text-sm hover:bg-background-components-hover focus-visible:shadow-button-neutral-focus focus-visible:outline-none"
+          type="button"
+        >
+          <Text as="span" size="sm" bold>
+            Acme
+          </Text>
+          <Icon name="arrowDownSLine" className="size-14 text-foreground-neutral-muted" />
+        </button>
+        <span className="text-foreground-neutral-disabled text-lg">/</span>
+        <button
+          className="flex h-28 min-w-0 items-center gap-4 rounded-6 px-6 text-foreground-neutral-base hover:bg-background-components-hover focus-visible:shadow-button-neutral-focus focus-visible:outline-none"
+          type="button"
+        >
+          <Code as="span" variant="label" className="truncate">
+            checkout-api
+          </Code>
+          <Icon name="arrowDownSLine" className="size-14 shrink-0 text-foreground-neutral-muted" />
+        </button>
+      </div>
+      <div className="flex items-center gap-6">
+        <button
+          className="hidden h-28 items-center gap-8 rounded-6 border border-border-neutral-base bg-background-components-base px-10 text-foreground-neutral-muted text-xs shadow-button-neutral hover:bg-background-components-hover focus-visible:shadow-button-neutral-focus focus-visible:outline-none sm:flex"
+          type="button"
+        >
+          <Icon name="searchLine" className="size-14" />
+          <span>Search runs, steps, commits...</span>
+          <Kbd className="h-18 min-w-18 px-3">/</Kbd>
+        </button>
+        <Button size="sm" variant="transparentMuted" iconLeft="bookOpen" aria-label="Docs" />
+        <Avatar size="sm" content="letters" fallback="TA" />
+      </div>
+    </header>
+  );
+}
--- a/libs/client/projects/src/pages/workflow-dashboard/components/step-detail-panel.tsx
+++ b/libs/client/projects/src/pages/workflow-dashboard/components/step-detail-panel.tsx
@@ -1,0 +1,372 @@
+import {Button, Code, cn, Icon, Text} from '@shipfox/react-ui';
+import {formatClock, formatDuration} from '../lib/workflow-dashboard-format.js';
+import {workflowStatusTextClass} from '../lib/workflow-dashboard-status.js';
+import type {
+  WorkflowDashboardAttempt,
+  WorkflowDashboardLogStream,
+  WorkflowDashboardRun,
+  WorkflowDashboardStep,
+  WorkflowDashboardViewModel,
+} from '../workflow-dashboard-types.js';
+import {WorkflowStatusBadge} from './status-badge.js';
+
+export type WorkflowDashboardViewMode = 'logs' | 'overview' | 'source';
+export type WorkflowDashboardLogFilter = 'all' | WorkflowDashboardLogStream;
+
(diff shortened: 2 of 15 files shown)


libs/client/projects/src/pages/workflow-dashboard/components/dashboard-shell.tsx
noe-charmet, line 1: In terms of pure file location, we usually position our components in `src/components`, pages are in `src/pages` and read from `src/components`, utils would be in `src/utils` or `src/hooks` if react hooks.

libs/client/projects/src/pages/workflow-dashboard/workflow-run-dashboard.tsx
noe-charmet, line 1: This page has way too many component inlined in the file, it can be split up in multiple files

libs/client/projects/src/pages/workflow-dashboard/workflow-run-dashboard.tsx
```diff
@@ -0,0 +1,430 @@
...
+  const job = run.jobs[0];
+
+  if (!job) {
+    throw new Error(`Workflow dashboard run #${run.number} has no jobs.`);
+  }
+
+  return job;
+}
+
+export function WorkflowRunDashboard({
```
noe-charmet, line 65: There is a lot of business logic in this component. I would suggest
- Extracting some to hooks, making it easy to unit test
- Extract some of it to an external state store, we use zustand for this (example in this package `libs/client/auth/src/state`)
- Investigate if some of the state can be located lower in the tree, to a more specialized sub-section of components. Seems like this is mostly done to pass a lot of data up and down the tree.

libs/client/projects/src/pages/workflow-dashboard/workflow-dashboard-types.ts
noe-charmet, line 1: Looks like you want to define some local state entities, that usually goes somewhere in `src/state`.

libs/client/projects/src/pages/workflow-dashboard/workflow-dashboard-types.ts
```diff
@@ -0,0 +1,94 @@
+export type WorkflowDashboardStatus =
```
noe-charmet, line 1: To avoid drift, why not read this directly from the backend DTO ?

libs/client/projects/src/pages/workflow-dashboard/lib/workflow-dashboard-format.ts
noe-charmet, line 1: This seems to add another method for representing durations from earlier in this stack
https://github.com/ShipfoxHQ/shipfox/pull/199#discussion_r3419188417

I would suggest unifying all of this and adding it to `@shipfox/react-ui` for tooling first.

libs/client/projects/src/pages/workflow-dashboard/lib/workflow-dashboard-format.ts
```diff
@@ -0,0 +1,31 @@
...
+  const minutes = Math.floor(rounded / 60);
+  const remainingSeconds = rounded % 60;
+  if (minutes < 60) return `${minutes}m${String(remainingSeconds).padStart(2, '0')}s`;
+  const hours = Math.floor(minutes / 60);
+  const remainingMinutes = minutes % 60;
+  return `${hours}h${String(remainingMinutes).padStart(2, '0')}m`;
+}
+
+export function formatClock(iso: string): string {
+  return new Date(iso).toLocaleTimeString('en-GB', {
```
noe-charmet, line 14: Why would we impose `en-GB` to the whole universe ?
If we leave `undefined` it'll adapt to the user's system language, and our Japanese users will be happy to have `一五時三六分`

libs/client/projects/src/pages/workflow-dashboard/lib/workflow-dashboard-format.ts
```diff
@@ -0,0 +1,31 @@
...
+    hour: '2-digit',
+    hour12: false,
+    minute: '2-digit',
+    second: '2-digit',
+  });
+}
+
+export function formatRelativeTime(iso: string, nowIso?: string): string {
```
noe-charmet, line 22: We can use [formatRelative](https://date-fns.org/v4.4.0/docs/formatRelative) from date-fns

libs/client/projects/src/pages/workflow-dashboard/lib/workflow-dashboard-status.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export function workflowStatusVariant(status: WorkflowDashboardStatus | string): BadgeVariant {
+  if (status === 'succeeded') return 'success';
+  if (status === 'failed' || status === 'timed-out') return 'error';
+  if (status === 'running') return 'info';
+  if (status === 'awaiting-runner') return 'warning';
+  return 'neutral';
+}
+
```
noe-charmet, line 27: Nitpick, but I feel we could have used a typed HashMap here to store the values and make things a bit more easy on the comprehension and readability.

```typescript
const workflowStatusVariant: Record<WorkflowDashboardStatus, BadgeVariant> = {
  succeeded: 'success',
  failed: 'error',
  'timed-out': 'error';
  // ...
}

// Inside the component
const statusVariant = workflowStatusVariant[status] ?? 'neutral'
```

libs/client/projects/src/pages/workflow-dashboard/lib/workflow-dashboard-status.ts
noe-charmet, line 1: Since these are exclusively used within `WorkflowStatusBadge`, maybe that we should co-locate them ?
You can have `workflow-status-badge.hooks.ts` or `workflow-status-badge.const.ts` right beside it.

libs/client/projects/src/pages/workflow-dashboard/workflow-run-dashboard.tsx
noe-charmet, line 65: This page is never mounted anywhere, how to you properly test this PR ?

libs/client/projects/src/pages/workflow-dashboard/components/dashboard-shell.tsx
```diff
@@ -0,0 +1,62 @@
...
+  );
+}
+
+function WorkflowDashboardTopNav() {
+  return (
+    <header className="flex items-center justify-between gap-12 border-border-neutral-base border-b bg-background-components-base px-16">
+      <div className="flex min-w-0 items-center gap-8">
+        <Logo className="h-18 w-auto shrink-0" variant="wordmark" />
+        <span className="text-foreground-neutral-disabled text-lg">/</span>
+        <button
```
noe-charmet, line 28: What prevents the usage of our `Button` component here ? And elsewhere in this PR. Do we need to add new variants or extend it ?

libs/client/projects/src/pages/workflow-dashboard/workflow-run-dashboard.tsx
noe-charmet, line 65: `WorkflowRunPage` would be a better name IMO. We already are inside a "dashboard" application, so you're not actually making a new dashboard.

libs/client/projects/src/pages/workflow-dashboard/components/step-detail-panel.tsx
noe-charmet, line 1: There are a lot of different features in this file. I would suggest splitting them up entirely, viewing Logs is not the same thing as viewing source files, both in terms of features and state management.

I would also vouch to maybe not have designs for non-existing features until we actually build them (ex: logs, restart...)

libs/client/projects/src/pages/workflow-dashboard/workflow-run-dashboard.tsx
```diff
@@ -0,0 +1,430 @@
...
+  onLogQueryChange: (query: string) => void;
+  onViewModeChange: (mode: WorkflowDashboardViewMode) => void;
+  viewMode: WorkflowDashboardViewMode;
+}) {
+  return (
+    <div className="flex flex-wrap items-center gap-8">
+      {viewMode === 'logs' && (
+        <label className="flex h-28 items-center gap-6 rounded-6 border border-border-neutral-base bg-background-field-base px-8 text-foreground-neutral-muted">
+          <Icon name="searchLine" className="size-14" />
+          <input
```
noe-charmet, line 309: Something prevents us from using our UI library's `Input` do we need to extend it to support some use cases ? Same for all other inputs in this PR

noe-charmet:
There are some good basis, and I feel a will to make things correctly.
However in all honesty all of this is quite a big mess, and it will take quite some effort to align the code with our standards.
I would suggest splitting work up in multiple independent pieces that can be more easy to reason around, I'd be happy to pair on getting this going. This is a big-bang change which introduces a huge feature at once, but would have a way better output by splitting this in multiple independent parts (ex: UI library extension, runs search & navigation, jobs graph, steps vizualisation, integration within the existing dashboard...).

ThierryAbalea:

PR #208, opened 2026-06-16
[client-projects] Add workflow run summary component

ThierryAbalea:
Add a compact workflow run summary component for the run detail experience.

The component renders the selected run's short identifier, workflow name, status, trigger source/event, neutral payload shape metadata, and created/updated relative timestamps from the existing workflow run DTO. It is intentionally component-level and integration-ready: no route composition, logs, duration, or run-control actions are introduced.

Validation covers failed, running, succeeded, and missing-trigger states with DTO-backed fixtures, pure mapper tests, and DOM render coverage. The isolated preview was rendered at desktop size and compared against the run summary region in the workflow run overview designs; real run-detail data validation remains part of the ENG-461/final integration path.

No changeset is included because @shipfox/client-projects is private.

Closes ENG-453

---
## Summary by cubic
Replaces the old run header with a compact `WorkflowRunSummary` on the Workflow Run page (ENG-453). It shows the run name with a status badge, a copyable short/long run id, a trigger label with icon, and a single “Triggered” time; the jobs graph stays below.

- **New Features**
  - Added `WorkflowRunSummary` mapping `RunResponseDto` to a view model; shows status next to the name and a relative “Triggered” time; supports pending/running/succeeded/failed/cancelled.
  - New `Identifier` control for the run id (shows 8-char short id, copies full id; keyboard-accessible with tooltip feedback for copied/failed).
  - Trigger label uses shared “source / event,” hides when empty, shows a source icon, with tooltip for long labels.

- **Refactors**
  - Replaced `WorkflowRunHeader` with `WorkflowRunSummary`; removed the old header and stories; updated run-view to use the summary and expanded tests.
  - Extracted `runTriggerLabel` to `workflow-run-display.ts` and reused it in the runs list; updated the loading skeleton to mirror the summary; added model/component tests; added `@shipfox/client-triggers` for trigger icons.

<sup>Written for commit eb6667646914ce0fca8b842455c77850a10a9a78. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ThierryAbalea:

libs/client/workflows/src/components/workflow-run-summary/workflow-run-summary.tsx
```diff
@@ -0,0 +1,154 @@
...
+          <Text as="span" size="xs" className="inline-flex items-center gap-4">
+            Created
+            <RelativeTime value={model.createdAt} className="font-code text-xs leading-20" />
+          </Text>
+          <Text as="span" size="xs" className="inline-flex items-center gap-4">
+            Updated
+            <RelativeTime value={model.updatedAt} className="font-code text-xs leading-20" />
+          </Text>
+        </div>
+      </div>
```
noe-charmet: I don't think having both updated and created is useful and it creates more noise.
Maybe that we could Replace it with "Triggered X ago" which would base itself on the createdAt property.

libs/client/workflows/src/components/workflow-run-summary/workflow-run-summary.tsx
```diff
@@ -0,0 +1,154 @@
...
+        <div className="flex min-w-0 items-center gap-8">
+          <Dot variant={model.status.dot} ripple={run.status === 'running'} className="size-8" />
+          <Header id={headingId} variant="h3" className="min-w-0 truncate">
+            {model.name}
+          </Header>
+        </div>
+
+        <Badge variant={model.status.badge} iconLeft={model.status.icon} size="xs">
+          {model.status.label}
+        </Badge>
```
noe-charmet: I would suggest for a more compact and efficient data
- Remove the dot (it's repeated information in the badge)
- Move the badge to first position. It's fixed length while the name isn't this means it will not move around based on the name length.

libs/client/workflows/src/components/workflow-run-summary/workflow-run-summary.tsx
```diff
@@ -0,0 +1,154 @@
...
+            Updated
+            <RelativeTime value={model.updatedAt} className="font-code text-xs leading-20" />
+          </Text>
+        </div>
+      </div>
+    </section>
+  );
+}
+
+function RunId({id, shortId}: {id: string; shortId: string}) {
```
noe-charmet: I would suggest extracting this to its own component as it seems to be not specific to this summary.
It could be more generic and be a `Identifier` component with `display` and `value` properties.

libs/client/workflows/src/components/workflow-run-summary/workflow-run-summary-model.test.ts
noe-charmet, line 1: To refactor in a follow-up PR, but I think we should extract a business model of the different workflows and do the mapping DTO -> model way earlier in the state. We probably want all our components to manipulate a single domain entity and not have each component do some local re-mapping like this.

libs/client/workflows/src/components/workflow-run-summary/workflow-run-summary.tsx
```diff
@@ -0,0 +1,154 @@
...
+        </Badge>
+
+        <span
+          aria-hidden="true"
+          className="hidden h-20 w-px shrink-0 bg-border-neutral-base sm:block"
+        />
+
+        <RunId id={model.id} shortId={model.shortId} />
+
+        {model.triggerLabel ? (
```
noe-charmet, line 53: We now have `TriggerSourceIcon` that exists, you probably want to add it beside the trigger name

noe-charmet:
The long name / trigger stories show some possible improvements
- On the long name one, the name is actually not long enough to trigger truncation/reflow
- On the long trigger one, we see the time reflow to a second line. I believe we'd rather truncate the text with an ellipsis + tooltip than have a reflow.

ThierryAbalea:
@noe-charmet handled in the latest push.

- Made the long-name story long enough to exercise truncation/reflow.
- Kept the triggered timestamp on one line.
- Made long trigger metadata truncate with ellipsis and expose the full value in a tooltip.

noe-charmet (dismissed):
Thank you for all the changes, looks great !
Let you rebase and update with the latest main before merging, we slightly changed the badge component usage.

ThierryAbalea:
Temporarily closing and reopening to retrigger PR checks.

PR #210, opened 2026-06-16
[client-workflows] Rework workflow run step overview (inline + Source panel)

ThierryAbalea:
Reworks the Workflow Run Page step overview to the updated Claude design and to live with the rest of the run-page components.

## What changed in the design

The expanded Step (attempt) view no longer switches between **Overview / Logs / Source** sub-views with tabs. There is a single expanded view that contains, top to bottom:

- the step **overview** (this PR),
- inline **logs** (a later PR, not here),
- a **Source** button that opens the run's workflow source in the existing right-side source panel (ENG-584), scrolled to and highlighting the step's source lines.

## What changed in the code

- The standalone `panel`/`inline` variant component in `@shipfox/client-projects` is replaced by a single **inline** `WorkflowStepOverview` in `@shipfox/client-workflows`, rendered through the existing `WorkflowStepList` `renderExpandedStep` slot. The old `client-projects` component, model, fixtures, stories, and tests are removed.
- Drops the `variant` prop, the panel chrome (`min-h-560`, outer card), and the "Select a step" empty state. The expanded row is the frame.
- Result/root-cause summary maps from the **typed** `gate_result` / `restart_result` contract (ENG-458, now on `main`), not opaque blobs.
- Adds a **Source** affordance shown only when the run has a `source_snapshot` and the step has a `source_location`. It opens the page-level `WorkflowSourcePanel` focused on the step's lines; the focus state is owned by `WorkflowRunView`.

## Boundaries

- **In-panel line highlight / auto-scroll rendering is out of scope** — that is ENG-460. This PR only threads the step's `source_location` into the panel and opens it; until ENG-460 lands the button opens the panel with the full source (graceful degradation).
- No logs, no restart/re-run actions, no duration, no backend/DTO changes (`source_location` and typed gate/restart already exist on `main`).

## Notes

This is a complete redo of the earlier `client-projects` implementation, not an incremental commit on top of it. Plan and updated scope: ENG-456.

Closes ENG-456


ThierryAbalea:

PR #211, opened 2026-06-16
[client-workflows] Add workflow jobs graph

ThierryAbalea:
## Summary

- Adds `WorkflowJobsGraph` under `@shipfox/client-workflows` and mounts it on the real `WorkflowRunView` path for the workflow run detail page.
- Builds a DTO-backed 2-D DAG from job-name dependencies, with trigger edges, dependency-depth columns, position-ordered parallel rows, and a softened selected-node highlight.
- Aligns job status handling with the runtime-produced states: `pending`, `running`, `succeeded`, `failed`, and `cancelled`; removes synthetic dependency-risk/manual/waiting UI states.
- Covers visible graph states with Storybook/Argos stories: single job, linear wide graph, parallel tall graph, branch/join, two running jobs, failed dependency cancelling downstream, empty state, selected state, and large wide+tall graph.
- Adds pure node tests for graph/status mapping plus DOM integration tests for selection, empty state, downstream cancellation, long names, and the mounted run detail view.

## Validation

- `mise exec -- turbo test --filter=@shipfox/client-workflows --filter=@shipfox/api-workflows`
- `mise exec -- turbo type --filter=@shipfox/client-workflows --filter=@shipfox/api-workflows`
- `mise exec -- turbo check --filter=@shipfox/client-workflows --filter=@shipfox/api-workflows`
- `CI=true mise exec -- turbo build --filter '...[origin/main]'`
- Storybook browser sanity check against the graph stories after consult fixes: parallel ordering, branch/join render, region labelling, and running-state icon noise.

Refs ENG-454


ThierryAbalea:

libs/client/workflows/src/components/workflow-status/status-visuals.ts
noe-charmet, line 1: Since these will be the same accross every single place in this repo, should we de-duplicate them from `/libs/client/workflows/src/components/workflow-runs-list/status-visuals.ts` ?

noe-charmet: We could also extract some generic `StatusBadge` / `StatusDot` components as we'll probably de using them quite often in this module.

libs/client/workflows/src/components/workflow-jobs-graph/workflow-jobs-graph-content.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  if (model.nodes.length === 0) {
+    return (
+      <div className="flex min-h-160 flex-col items-center justify-center gap-6 px-16 py-24">
+        <Text size="sm" bold className="text-foreground-neutral-base">
+          No jobs yet
+        </Text>
+        <Text size="xs" className="text-foreground-neutral-muted">
+          This run has not materialized jobs.
+        </Text>
```
noe-charmet: For UI consistency you can use our EmptyState component

libs/client/workflows/src/components/workflow-jobs-graph/workflow-job-node.tsx
noe-charmet, line 1: Would be great to have stories for WorkflowJobNode in isolation. At least to have all the different variants in a single screen for UI regression tests.

libs/client/workflows/src/components/workflow-jobs-graph/workflow-job-node.tsx
```diff
@@ -0,0 +1,91 @@
...
+          {visual.label}
+        </Badge>
+        {node.dependencies.length > 0 ? (
+          <Code
+            variant="label"
+            className="min-w-0 flex-1 truncate text-foreground-neutral-muted"
+            title={`needs ${node.dependencies.join(', ')}`}
+          >
+            needs {node.dependencies.join(', ')}
+          </Code>
```
noe-charmet: I tend to think that the badge is creating noise here on a surface for which we have little space. Would the Dot be enough to represent current state in this situation ?

noe-charmet: Is there a specific reason to use monospace fonts with `Code` component here ?
It would seem to me that this surface is representing regular text and would better fit in the general design with our usual text fonts.

libs/client/workflows/src/components/workflow-jobs-graph/workflow-job-node.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Dot variant={visual.dot} ripple={node.statusKind === 'running'} />
+        <span className="sr-only">{visual.label}</span>
+        <Code
+          variant="label"
+          bold
+          className="truncate text-foreground-neutral-base"
+          title={node.label}
+        >
+          {node.label}
+        </Code>
```
noe-charmet: Since this can be truncated, could we add a Tooltip to display complete string on hover ?

libs/client/workflows/src/components/workflow-jobs-graph/workflow-job-node.tsx
noe-charmet: IMO adding all these strings like this will quickly overflow and be hard to read for our end-users.

If we think that the graph is not enough to convey the dependency need, I'd simplify this with "Waiting for X jobs to finish". We could also potentially have job names on hover, but I really believe we'll be repeating what the graph itself is saying again.

libs/client/workflows/src/components/workflow-jobs-graph/workflow-jobs-graph-view.tsx
```diff
@@ -0,0 +1,46 @@
...
+          </Text>
+          <Text size="xs" className="text-foreground-neutral-muted">
+            {model.nodes.length} {model.nodes.length === 1 ? 'job' : 'jobs'}
+          </Text>
+        </div>
+        {model.warnings.length > 0 ? (
+          <Text size="xs" className="shrink-0 text-foreground-neutral-muted">
+            Large graph
+          </Text>
+        ) : null}
```
noe-charmet: What's the goal with this warning ? What information do we convey to the user compared to a smaller graph ?

BTW there is no story covering this case

libs/client/workflows/src/components/workflow-jobs-graph/workflow-jobs-graph.stories.tsx
noe-charmet, line 1: One case that is not covered is when you have jobs in this dependency situation.

```mermaid
flowchart LR
    A[Job A] --> B[Job B] --> C[Job C]
    D[Job D] --> C
```

noe-charmet: Note that the mermaid representation is wrong, as we'd expect A and D to be vertically aligned

libs/client/workflows/src/components/workflow-jobs-graph/workflow-job-node.tsx
noe-charmet, line 1: Not sure if it would be overkill or not for our usage, but have you evaluated using a dedicated flow charting library like [React Flow](https://reactflow.dev/) ?

noe-charmet:
Thanks for taking time for this one 💪 It really is a great step-up in quality of render.
I left some comments mostly on some UI/UX improvements / suggestions / discussions, and a few technical/testing challenges.

libs/client/workflows/src/components/workflow-jobs-graph/workflow-job-node.tsx
ThierryAbalea: I’d defer React Flow for now because this slice is a compact read-only graph, and revisit it if we need pan/zoom, minimap, fit-to-view, editing, or edge-routing that becomes too costly to maintain locally.

PR #216, opened 2026-06-17
Propagate projectId end-to-end via the job lease

noe-charmet:
Threads `projectId` from the workflow run through the runner job tables into the signed job-lease token, so a claimed job's lease now carries `(workspaceId, projectId, runId, jobId)` instead of only `(workspaceId, runId, jobId, runnerTokenId)`.

## Why

Log-ingest will need to authorize and stamp per-project on every access. With `projectId` absent from the lease and the runner job rows, that means joining back to `workflows` at ingest time. Putting `projectId` on the signed lease makes the identifier set self-contained: a claimed job's lease names its project, determined at mint time from server state.

`projectId` is functionally determined by `jobId` via immutable `workflows` FKs (`workflow_runs.project_id` → run → job). So the signed claim is denormalization for self-contained authorization and signed defense-in-depth, **not** a fix for an existing forgery path. The runtime value is belt-and-suspenders, and this PR intentionally does not oversell it.

## What changed

- **auth / auth-dto** — `projectId` added to `jobLeaseTokenClaimsSchema` and signed by `issueJobLeaseToken`. It surfaces on `LeasedJobContext` automatically (the context is the claims type).
- **runners** — `project_id NOT NULL` on `runners_pending_jobs` and `runners_running_jobs`; threaded through `scheduleJob` → `claimPendingJob` (carried pending → running on the claim) → `claimJob` → lease. The migration is hand-edited into the initial migration since the DB is recreated.
- **workflows** — `loadRunDag` surfaces `run.projectId`, and `launchJobs` now sources the whole `{workspaceId, projectId, runId}` identity tuple from the loaded run (rather than mixing the run-start input with the loaded `projectId`), threading it through `JobOrchestrationInput` → `enqueueJobForRunner`.

## Scope / review notes

- **Producer-side only.** The `attempt_streams` consumer (stamp `run_id`/`project_id` at create, assert against the lease, `LeaseStreamMismatchError`) is deferred to the logs-module PR — that module does not exist in the repo yet. The deferred consumer work will be tracked under a separate ticket.
- **Immutability assumption.** This relies on a run never being re-parented to a different project. That holds today; the short-lived lease (and recreated rows) keep any future staleness window tiny.
- **Source-of-truth coherence.** `launchJobs` was changed to source `workspaceId`/`runId` from the loaded dag too (not just `projectId`), so the whole signed tuple is server-derived.

Closes ENG-467

---
## Summary by cubic
Propagates `projectId` end-to-end and adds it to the signed runner job lease so each claimed job now includes `{workspaceId, projectId, runId, jobId}`. This enables per‑project log‑ingest authorization without joining back to workflows. Closes ENG-467.

- **New Features**
  - `@shipfox/api-auth-dto`, `@shipfox/api-auth`: added `projectId` to `jobLeaseTokenClaimsSchema`; `issueJobLeaseToken` now signs it.
  - `@shipfox/api-runners`: added `project_id` to `runners_pending_jobs` and `runners_running_jobs`; threaded `projectId` through `scheduleJob` → `claimPendingJob` → `claimJob` into the lease token.
  - `@shipfox/api-workflows`: `loadRunDag` now returns `{workspaceId, projectId, runId}`; `runOrchestration` and `enqueueJobForRunner` use these run‑derived ids when scheduling.

- **Bug Fixes**
  - `@shipfox/api-runners`: synced Drizzle meta snapshot with `project_id` to match the initial SQL and prevent drift/redundant migrations.

<sup>Written for commit d37cad73a2caec25034771d03968a60d8ee3345c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libs/api/auth-dto/src/schemas/job-lease-token.ts
+++ b/libs/api/auth-dto/src/schemas/job-lease-token.ts
@@ -10,6 +10,7 @@ export const JOB_LEASE_TOKEN_AUDIENCE = 'runner-job-lease';
 export const jobLeaseTokenClaimsSchema = z.object({
   jobId: z.string().uuid(),
   runId: z.string().uuid(),
+  projectId: z.string().uuid(),
   workspaceId: z.string().uuid(),
   runnerTokenId: z.string().uuid(),
   aud: z.literal(JOB_LEASE_TOKEN_AUDIENCE),
--- a/libs/api/workflows/src/temporal/workflows/run-orchestration.ts
+++ b/libs/api/workflows/src/temporal/workflows/run-orchestration.ts
@@ -4,7 +4,7 @@ import type {RuntimeCompletionStatus} from '#core/entities/runtime-dag.js';
 import {scheduleRuntimeDag} from '#core/workflow-runtime/index.js';
 
 import type {createOrchestrationActivities} from '../activities/index.js';
-import type {DagJob} from '../activities/orchestration-activities.js';
+import type {DagJob, RunDag} from '../activities/orchestration-activities.js';
 import {jobOrchestration} from './job-orchestration.js';
 
 const {loadRunDag, setRunStatus, setJobStatus} = proxyActivities<
@@ -47,7 +47,7 @@ export async function runOrchestration(input: RunOrchestrationInput): Promise<vo
     const jobsToStart = commands.flatMap((command) =>
       command.kind === 'start-job' ? [command.job] : [],
     );
-    const results = await launchJobs(jobsToStart, input, jobVersions);
+    const results = await launchJobs(jobsToStart, dag, jobVersions);
     for (const [job, result] of jobsToStart.map((j, i) => [j, results[i]] as const)) {
       if (!result) continue;
       completed.set(job.name, result.status);
@@ -83,7 +83,7 @@ interface LaunchResult {
 
 function launchJobs(
   jobs: DagJob[],
-  input: RunOrchestrationInput,
+  run: RunDag,
   jobVersions: Map<string, number>,
 ): Promise<LaunchResult[]> {
   return Promise.all(
@@ -92,9 +92,10 @@ function launchJobs(
         workflowId: `job:${job.id}`,
         args: [
           {
-            workspaceId: input.workspaceId,
+            workspaceId: run.workspaceId,
+            projectId: run.projectId,
             jobId: job.id,
-            runId: input.runId,
+            runId: run.runId,
             jobVersion: jobVersions.get(job.id) ?? job.version,
           },
         ],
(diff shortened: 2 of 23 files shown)


libs/api/auth-dto/src/schemas/job-lease-token.ts
```diff
@@ -10,6 +10,7 @@ export const JOB_LEASE_TOKEN_AUDIENCE = 'runner-job-lease';
 export const jobLeaseTokenClaimsSchema = z.object({
   jobId: z.string().uuid(),
   runId: z.string().uuid(),
+  projectId: z.string().uuid(),
```
ThierryAbalea, line 13: Small rollout check: making `projectId` required on the verify path means leases minted before this deploy will fail auth until they expire (`AUTH_JOB_LEASE_TOKEN_EXPIRES_IN` defaults to 90m). If we can guarantee no in-flight runner jobs during deploy, fine; otherwise could we tolerate missing `projectId` on verify for one release while still requiring it on issue, or document the drain/rotation plan?

libs/api/workflows/src/temporal/workflows/run-orchestration.ts
```diff
@@ -92,9 +92,10 @@ function launchJobs(
         workflowId: `job:${job.id}`,
         args: [
           {
-            workspaceId: input.workspaceId,
+            workspaceId: run.workspaceId,
+            projectId: run.projectId,
```
ThierryAbalea, line 96: Related deploy-straddling check: `loadRunDag()` activity results recorded before this change do not include `projectId`, but new replays use `run.projectId` when starting later child jobs. If runs can stay active across deploys, could we either drain them or tolerate/reload the project id during the transition?

libs/api/auth-dto/src/schemas/job-lease-token.ts
noe-charmet: Thanks Claude, but we never deployed any of this code anywhere 👍

libs/api/workflows/src/temporal/workflows/run-orchestration.ts
noe-charmet: Thanks Claude, but we never deployed any of this code anywhere 👍

PR #235, opened 2026-06-19
[api/integration] Move source-event translation into the integration module

noe-charmet:
Source-control providers now own the translation of their webhooks into a typed, provider-agnostic domain event, instead of the `projects` module decoding GitHub-specific payloads.

## Why

`projects`' subscriber read the generic `INTEGRATION_EVENT_RECEIVED` envelope and hard-coded provider knowledge (`source === 'github'`, `event === 'push'`, a cast to `GithubPushPayload`, `isDefaultBranch`). That put "how to interpret each integration's push webhook" inside the projects domain: every new provider (GitLab, Bitbucket) would grow a `(source, event)` switch there, and every other domain consumer that wanted to react to a commit would repeat the same decoding.

The key constraint is that `triggers` also consumes `INTEGRATION_EVENT_RECEIVED`, but on purpose: it is a source-agnostic pass-through that fans `(source, event, payload)` out to user-configured workspace subscriptions. So the fix is not to remove the generic envelope, but to give *domain* consumers a typed event while keeping the raw stream for the pass-through consumer.

## What changed

- **`core-dto`**: renamed `GithubPushPayload` -> `SourcePushPayload` (provider-agnostic); added `INTEGRATION_SOURCE_COMMIT_PUSHED` + `IntegrationSourceCommitPushedEvent` (nests the push) and a `PublishSourcePushFn` port (kept in `core-dto` so providers don't depend on `core`, avoiding a cycle).
- **`integration/core`**: new `publishSourcePush` does the delivery-dedup once and writes **both** outbox rows — the generic envelope (for `triggers`) and the typed event (for `projects`) — in one transaction. It is transaction-only so the three writes can't partially commit.
- **`integration/github`**: `handleGithubPush` builds a `SourcePushPayload` and calls `publishSourcePush`; the injected dependency is swapped through the route -> factory -> core seam. Branch deletions (all-zero `after` SHA) are dropped at the normalizer rather than surfacing as observed commits.
- **`projects`**: `onSourceCommitPushed` (renamed from `onIntegrationEventReceived`) reads the typed event, applies the default-branch policy, resolves the project, dedups, and emits `PROJECT_SOURCE_COMMIT_OBSERVED` — no provider knowledge. `triggers` is unchanged.

## Notes for review

- `defaultBranch` filtering stays in `projects` (a consumer policy), so the integration event carries every push and a future consumer (e.g. PR checks) can react to non-default branches.
- Tests are per-layer: the "both rows / duplicate writes neither" assertions live in the `integration/core` publisher test (real DB), since the github route test mocks the publisher; `projects` gains a module-registration assertion in place of the removed non-github/non-push filter tests.
- Out of scope (follow-ups): a typed Sentry event when a domain consumer needs one, and making `writeOutboxEvent` generic over the event map repo-wide so consumer-side payload drift is a compile error too.

---
## Summary by cubic
Moved webhook translation into the integration module so providers emit typed, provider-agnostic events. A transactional publisher writes both the typed `INTEGRATION_SOURCE_COMMIT_PUSHED` and the generic `INTEGRATION_EVENT_RECEIVED`; `projects` now consumes the typed event while `triggers` still reads the generic stream.

- New Features
  - Added `INTEGRATION_SOURCE_COMMIT_PUSHED` and `publishSourcePush` to dedup by delivery and write both events in one transaction.
  - Introduced `SourcePushPayload`; GitHub normalizer drops branch-deletion pushes (all-zero SHA) so neither the typed event nor the generic envelope is emitted.

- Refactors
  - `projects` replaces provider-specific decoding with `onSourceCommitPushed`; default-branch filtering remains in `projects`.
  - Renamed `GithubPushPayload` to `SourcePushPayload`; tests updated per layer.

<sup>Written for commit c5d79e8cbe9d3b3cc764833332a3eb5d97716bff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/calm-foxes-share.md
+++ b/.changeset/calm-foxes-share.md
@@ -1,0 +1,8 @@
+---
+"@shipfox/api-integration-core-dto": minor
+"@shipfox/api-integration-core": minor
+"@shipfox/api-integration-github": patch
+"@shipfox/api-projects": patch
+---
+
+Moves source-event translation to the integration module: source-control providers emit a typed, provider-agnostic `INTEGRATION_SOURCE_COMMIT_PUSHED` event via one transactional publisher, projects subscribes to it instead of decoding GitHub payloads, and branch-deletion pushes are dropped at the source.
--- a/libs/api/integration/core-dto/src/events.ts
+++ b/libs/api/integration/core-dto/src/events.ts
@@ -10,12 +10,30 @@ export interface IntegrationEventReceivedEvent {
   payload: unknown;
 }
 
-export interface GithubPushPayload {
+// A source-control push, normalized by the producing provider. Carried both as the
+// generic `INTEGRATION_EVENT_RECEIVED` envelope payload (consumed opaquely by triggers)
+// and nested inside `INTEGRATION_SOURCE_COMMIT_PUSHED` (consumed by domain modules).
+export interface SourcePushPayload {
   externalRepositoryId: string;
   ref: string;
   headCommitSha: string;
   defaultBranch: string;
   isDefaultBranch: boolean;
+}
+
+export const INTEGRATION_SOURCE_COMMIT_PUSHED =
+  'integrations.source_control.commit_pushed' as const;
+
+// Typed, provider-agnostic source-control event. The producing provider owns the
+// translation from its raw webhook into this shape, so domain consumers never decode
+// provider payloads. `isDefaultBranch` is a fact; the branch policy lives in the consumer.
+export interface IntegrationSourceCommitPushedEvent {
+  provider: string;
+  workspaceId: string;
+  connectionId: string;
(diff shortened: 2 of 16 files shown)


noe-charmet:
Will merge this one as it's a small scope refactor and required to continue some other tasks

PR #237, opened 2026-06-19
[api/integrations] Move Sentry maintenance cron to the sentry module

noe-charmet:
Moves the Sentry unclaimed-installation cleanup cron out of `@shipfox/api-integration-core` and into `@shipfox/api-integration-sentry`. The sentry package now owns its own Temporal maintenance worker end-to-end (workflow, activity, schedule, retention), and `core` aggregates per-integration workers from enabled providers instead of special-casing Sentry.

## Why

`core` is both the generic integration kernel and the composition root, which let Sentry-specific business logic leak into it: the prune workflow, its activity (with an `await import` hack to dodge config loading when Sentry was disabled), and the retention constant all lived in `core/src/temporal/`. This restores clean ownership and keeps the integration dependency graph a tree: provider packages depend only on `*-core-dto`, never on `core`; `core` depends on providers solely to compose them.

It also matches the established convention where every other module (`runners`, `projects`, `workflows`, `dispatcher`, `definitions`) owns its own Temporal worker pointing at its own `dist/temporal/workflows/index.js`. Integration `core` was the only outlier.

## What changed

- `sentry` gains `src/temporal/` and exports `createSentryMaintenanceWorker(): ModuleWorker` on its own `integrations-sentry-maintenance` queue; the activity calls the in-package prune directly (the dynamic import is gone).
- `core` collects `ModuleWorker[]` from enabled providers and assembles `[coreMaintenanceWorker, ...providerWorkers]`; the generic webhook-delivery prune stays in `core`.
- The unclaimed-installation retention window is now an operator-facing config var, `SENTRY_UNCLAIMED_INSTALLATION_RETENTION_DAYS` (default 7), replacing the hardcoded constant.

## Alternatives considered

A single shared maintenance queue where `core` re-exports each integration's workflow into its bundle. Rejected: it introduces a pattern no other module uses and keeps per-integration coupling in `core`'s barrels. The per-integration worker matches convention and makes adding the next integration's cron a near-zero-change operation in `core`.

## Review notes

- `@temporalio/workflow` is a devDependency, matching `runners`/`projects`/`dispatcher` (which run their workers in production).
- Nothing is deployed yet, so there is no prior Temporal schedule to retire.

---
## Summary by cubic
Moved the Sentry unclaimed-installation cleanup cron from `@shipfox/api-integration-core` to `@shipfox/api-integration-sentry`, which now owns its Temporal maintenance worker and schedule. Retention is configured via `SENTRY_UNCLAIMED_INSTALLATION_RETENTION_DAYS` (default 7, minimum 1); startup fails on a smaller value.

- **Refactors**
  - `sentry` exports `createSentryMaintenanceWorker()` on queue `integrations-sentry-maintenance` with its own workflow and activity.
  - `core` aggregates per-integration `ModuleWorker`s; webhook-delivery prune remains in `core`.
  - Removed Sentry-specific activity/constant from `core` and dropped unused `sentry` root re-exports.

- **Migration**
  - Deploy the Sentry maintenance worker for task queue `integrations-sentry-maintenance`.
  - Set `SENTRY_UNCLAIMED_INSTALLATION_RETENTION_DAYS` to at least 1 (defaults to 7).
  - No existing Temporal schedule to retire.

<sup>Written for commit b8ecac8c78def34fae6f55ed711a174dc515a0cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/237?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


noe-charmet:
Merging as its a small scoped refactor

PR #239, opened 2026-06-19
Fix invisible keyboard focus rings

noe-charmet:
Fix invisible keyboard focus rings on project cards, integration tiles, and the user menu trigger.

These interactive surfaces removed the default outline but pointed at the non-existent `shadow-button-secondary-focus` token, leaving keyboard users without a visible focus state. This swaps them to the existing neutral button focus token and adds render-level regression assertions so the bad token does not come back.

Considered adding a secondary focus token, but the design system already defines `shadow-button-neutral-focus`, and PR #207 uses that same token for the nav logo fix. This keeps ENG-515 scoped to the remaining call sites.

Includes a patch changeset for the affected client packages.

Closes ENG-515

---
## Summary by cubic
Fix invisible keyboard focus rings on project cards, integration tiles, and the user menu by switching to the existing neutral focus token. Restores a clear focus state for keyboard users and adds tests to prevent regressions. Closes ENG-515.

- **Bug Fixes**
  - Replaced `focus-visible:shadow-button-secondary-focus` with `focus-visible:shadow-button-neutral-focus` on user menu, integration tiles, and project cards.
  - Added tests to assert the neutral focus class is present and the secondary token is not.
  - Published patch updates for `@shipfox/client-app-shell`, `@shipfox/client-integrations`, and `@shipfox/client-projects`.

<sup>Written for commit c0dd1f380c20ff7e763e1f0b884d8c219fca69bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/239?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/steady-links-focus.md
+++ b/.changeset/steady-links-focus.md
@@ -1,0 +1,7 @@
+---
+"@shipfox/client-app-shell": patch
+"@shipfox/client-integrations": patch
+"@shipfox/client-projects": patch
+---
+
+Fixes invisible keyboard focus rings on the user menu, integration tiles, and project cards by using the existing neutral button focus token.
--- a/libs/client/app-shell/src/components/user-menu.test.tsx
+++ b/libs/client/app-shell/src/components/user-menu.test.tsx
@@ -32,7 +32,11 @@ describe('UserMenu', () => {
       </ThemeProvider>,
     );
 
-    fireEvent.pointerDown(screen.getByRole('button', {name: 'User menu'}), {
+    const trigger = screen.getByRole('button', {name: 'User menu'});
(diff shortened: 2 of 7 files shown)


noe-charmet:
Will merge as it is a minimal UI fix

PR #241, opened 2026-06-19
Type the event-to-payload match for outbox subscribers

noe-charmet:
Outbox subscribers cast `event.payload as X` by hand in every handler, so a wrong event name or payload shape only surfaced at runtime. This adds a typed `subscriberFactory` in `@shipfox/node-module` that binds each event name to its payload type through the existing per-package event maps, the mirror image of the already-typed publish side (`writeOutboxEvent<TMap>`). Handlers now receive a typed `(payload, event)` and all 11 per-handler casts are gone.

Each module instantiates one factory bound to the intersection of the event maps it subscribes to, so a typo'd event name or a mismatched handler payload now fails to compile.

A module-private `Symbol` brands `ModuleSubscriber` so the factory is the only way to construct one. This was a deliberate call over leaving the helper opt-in: it stops a future raw `{event, handler}` literal from silently reintroducing an unchecked cast. The brand survives `.d.ts` emit (verified up front, since a type-only `unique symbol` would have broken it) and its blast radius is contained, as the five module declarations are the only places that construct subscribers.

Tests: a unit test for the factory (runtime forwarding plus `@ts-expect-error` guards that `turbo type` enforces), the five existing subscriber tests updated to the new call convention, and behavior tests backfilled for five handlers that previously had none.

### Review notes

- The payload cast still happens once inside the factory and is unchecked at runtime (the outbox stores JSONB). This is not a regression; the previous per-handler casts were equally unchecked. Validating the payload at the drain boundary is deferred to ENG-517.
- `@shipfox/node-module` is private, so the brand is an internal-only construction change and every call site is migrated in this PR.

Refs ENG-517

---
## Summary by cubic
Adds a typed `subscriberFactory` in `@shipfox/node-module` that binds outbox event names to payload types, removing per-handler casts and catching mismatches at compile time. Advances ENG-517; runtime payload validation at the drain boundary remains deferred.

- New Features
  - `subscriberFactory<TMap>()` returns subscribers typed by event maps; handlers receive `(payload, event)`.
  - `ModuleSubscriber` is branded with a private symbol so only the factory can construct it.
  - Centralizes the single payload cast at the factory boundary; no runtime behavior change.

- Refactors
  - Migrated subscribers in `@shipfox/api-definitions`, `@shipfox/api-projects`, `@shipfox/api-runners`, `@shipfox/api-triggers`, `@shipfox/api-workflows`, and `@shipfox/api-logs` to the factory; standardized handler signatures and removed casts.
  - Added a unit test for the factory and backfilled behavior tests; hardened idempotent swallow-path tests to assert Temporal is invoked once.
  - Added `@shipfox/vitest` and test scripts in `@shipfox/node-module`; updated the README example and documented the event-map disjointness constraint.

<sup>Written for commit d890ab07776ffda361c38a0ad51f7b5857e95e58. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/241?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/spry-falcons-gather.md
+++ b/.changeset/spry-falcons-gather.md
@@ -5,6 +5,7 @@
 "@shipfox/api-definitions": patch
 "@shipfox/api-projects": patch
 "@shipfox/api-runners": patch
+"@shipfox/api-logs": patch
 ---
 
 Adds a typed `subscriberFactory` that binds each outbox event name to its payload type at construction, so subscriber handlers receive a typed `(payload, event)` and the per-handler `event.payload as X` casts are gone; a private brand makes the factory the only way to build a module subscriber.
--- a/.changeset/spry-falcons-gather.md
+++ b/.changeset/spry-falcons-gather.md
@@ -1,0 +1,10 @@
+---
+"@shipfox/node-module": minor
+"@shipfox/api-triggers": patch
+"@shipfox/api-workflows": patch
+"@shipfox/api-definitions": patch
+"@shipfox/api-projects": patch
+"@shipfox/api-runners": patch
+---
+
+Adds a typed `subscriberFactory` that binds each outbox event name to its payload type at construction, so subscriber handlers receive a typed `(payload, event)` and the per-handler `event.payload as X` casts are gone; a private brand makes the factory the only way to build a module subscriber.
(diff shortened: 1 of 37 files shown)


noe-charmet:
Will merge this once CI green, to avoid too many conflicts in usage with other branches

PR #243, opened 2026-06-19
[api/integration] Re-sync debug workflows on each API startup

noe-charmet:
When the debug integration provider is enabled, its repositories are static fixtures with no webhook, so workflow definitions never re-sync after the initial bind. Editing the debug fixtures (or changing the definitions schema) left no easy way to re-run the sync.

On each API startup, when the debug provider is enabled, the API now emits one `INTEGRATION_SOURCE_COMMIT_PUSHED` for the debug `platform` repo on its main branch, for every active debug connection. That re-drives the existing definitions-sync pipeline (projects subscriber, then definitions, then Temporal), so the debug workflows are re-discovered and re-applied on every boot. It is a no-op when the debug provider is disabled.

Key decisions:

- Only the typed event is emitted, never the generic `INTEGRATION_EVENT_RECEIVED` envelope, so triggers never re-run `on_push` workflows on restart.
- The re-sync logic lives in the debug provider package (`libs/api/integration/debug`); core only contributes the two injected DB primitives (list debug connections, write the typed outbox row) and the boot trigger on `IntegrationsContext`.
- The boot hook is error-isolated (try/catch + log), so a re-sync failure never blocks API startup, matching the existing `startModuleWorkers` posture.
- `headCommitSha` is a stable synthetic constant (`debug-startup-resync`), so the same `definition-sync` workflow id re-runs each boot via Temporal's `ALLOW_DUPLICATE`; `deliveryId` is unique per emission for log traceability.

Review notes: unit coverage asserts the emission contract, including the zero-envelope guarantee and the boot error isolation. Not exercised in CI: the full cross-module chain and the Docker end-to-end (enable the provider, bind a project to `debug:platform`, restart, confirm a sync runs while no `on_push` run fires).

---
## Summary by cubic
Re-syncs debug workflow definitions on every API startup by emitting a typed commit-pushed event for the debug `platform` repo per active debug connection. No effect when the debug provider is disabled, and it never triggers `on_push` workflows.

- **New Features**
  - Adds generic provider `startupTasks` executed post-migrations via `integrations.runStartupTasks()`; the debug provider registers a re-sync task; no-op if the provider isn’t enabled.
  - Emits only `INTEGRATION_SOURCE_COMMIT_PUSHED` with `headCommitSha` `debug-startup-resync`, a unique `deliveryId`, and skips delivery dedup; never writes the generic envelope.
  - Logic lives in `@shipfox/api-integration-debug`; core injects `listIntegrationConnectionsByProvider` and `publishSourceCommitPushed`, filters to `active` connections, attempts every connection, and logs aggregated failures without blocking boot.

- **Dependencies**
  - Added `@shipfox/node-opentelemetry` for logging in `@shipfox/api-integration-core`.

<sup>Written for commit f535eb555850347dc2b1a7f01c397ef902eab9a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/debug-startup-resync.md
+++ b/.changeset/debug-startup-resync.md
@@ -1,0 +1,6 @@
+---
+"@shipfox/api-integration-debug": patch
+"@shipfox/api-integration-core": patch
+---
+
+Debug integration: on each API startup (when the debug provider is enabled), emit an `INTEGRATION_SOURCE_COMMIT_PUSHED` for the debug `platform` repo on its main branch, for every active debug connection. This forces a re-sync of the debug workflow definitions on every boot. Only the typed event is emitted (not the generic `INTEGRATION_EVENT_RECEIVED` envelope), so it never re-runs `on_push` workflows.
--- a/apps/api/src/core/run.ts
+++ b/apps/api/src/core/run.ts
@@ -49,6 +49,10 @@ export async function run(): Promise<void> {
       dispatcherModule,
     ],
   });
+  // Boot-time module tasks (post-migration). When the debug provider is enabled this forces
+  // a definitions re-sync of the debug fixtures; it is a no-op otherwise and never throws.
+  await integrations.runStartupTasks();
+
   const e2eAuth = config.E2E_ENABLED ? [createE2eAdminAuthMethod(config)] : [];
   const mountedE2eRoutes = createE2eRouteGroup(e2eRoutes, config);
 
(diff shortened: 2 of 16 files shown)


noe-charmet:
This only affects the debug provider, not critical, will merge

PR #244, opened 2026-06-19
[api/logs] Close log streams on the runner's end record or job termination

noe-charmet:
Adds stream close to the log module: a stream can now move out of `open` so it becomes eligible for compaction. This is the close behaviour of ENG-442; compaction (ENG-519) and retention deletion (ENG-520) are separate issues.

The log module (ENG-439) accepted chunks via the offset-CAS append endpoint but never moved a stream out of `open`, so nothing could be compacted and Postgres chunk volume was bounded only by retention.

Two ways a stream closes, both routed through one `closeStream` core that writes a single `logs.stream.closed` outbox event (the compaction trigger, handled by ENG-519):

- **Declared close** — a committed `end` record closes the stream inside the append transaction (the fast, common path). `end.total_bytes` is decoded payload bytes, a different axis from the raw-spool `committed_length`, so close triggers on *receiving* the record, not a byte comparison.
- **Timeout close** — a `WORKFLOWS_JOB_TERMINATED` subscriber (the single reliable job-terminal signal, from #236) arms a grace-then-close Temporal workflow that force-closes any stream the runner never ended itself. Machine death, budget cap, spool failure, and cancellation all leave a stream open with no `end`, and the runner is no liveness signal (a quiet step sends zero appends). It appends a `runner_lost` tombstone and marks the stream truncated.

A closed-stream guard drops later appends so compaction can never race a late chunk. Tombstones are stored as control chunks *without advancing* `committed_length`, keeping the offset-CAS axis equal to runner spool bytes (compaction orders by `seq`, so they still concatenate correctly).

Schema: adds `closed_at` plus three partial indexes. Since nothing is deployed yet, this folds into the initial migration in place rather than a new migration file.

**Review focus:** the exactly-once close — `closeStream`'s guarded `UPDATE … WHERE state='open'` is both the row lock and the idempotency gate, since the declared and timeout paths can race the same stream — and the tombstone no-advance invariant.

Fixes ENG-518
Refs ENG-442

### DIFF
--- a/.changeset/eng-518-logs-stream-close.md
+++ b/.changeset/eng-518-logs-stream-close.md
@@ -1,0 +1,5 @@
+---
+"@shipfox/api-logs": patch
+---
+
+Close a log stream so it becomes eligible for compaction. A committed `end` record declared-closes the stream inside the append transaction, and a `WORKFLOWS_JOB_TERMINATED` subscriber arms a grace-then-close Temporal workflow that force-closes any stream the runner never ended (appending a `runner_lost` tombstone and marking it truncated). Both paths route through one guarded close that writes a single `logs.stream.closed` outbox event, and a closed-stream guard drops later appends. Adds `closed_at` and three partial indexes to the stream table.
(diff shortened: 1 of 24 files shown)


noe-charmet:
Will merge this one to progress on log handling

PR #298, opened 2026-06-23
[client-logs] Add log record components for process and system logs

noe-charmet:
Adds `@shipfox/client-logs`, the presentational record components for the step-log read stream. They compose the `@shipfox/react-ui` log primitives into renderers for every process and system record (`output`, `group_start`/`group_end`, `end`, `gap`, `capped`, `runner_lost`); `agent_session` is left to the agent-sessions surface.

The react-ui log viewer ships only structural primitives ("four primitives, zero semantics"); record renderers are meant to be composed by the consuming client. This package is the forward-looking home for the full step-log viewer (records now; read-path hook, step-detail page, and agent-session later), mirroring the `@shipfox/client-workflows` per-package Storybook + Argos recipe.

Scope is UI only, no state management. The read-path data wiring and the step-detail page are out of scope (separate issues).

### What's here
- `buildLogTree` reconstructs the group tree from the flat record list: `group_end` closes the matching `group_id` (robust to a `group_start` dropped under gap/backlog pressure), record dispatch is an exhaustive switch, and each group node carries a precomputed `hasError` and subtree line count.
- `OutputLogRow` (stdout/stderr; stderr distinguished by a subtle left rule, not a background tint), `LogGroup` (collapsible with running/duration/incomplete affordances and an inset error bar), the four system markers, and the `LogView` dispatcher with an empty state and copy-to-clipboard.
- A package-local Storybook with a story per record type and an interleaved showcase, captured by Argos (build `client-logs`).

### Needs a careful look
`libs/api/logs-dto/src/schemas/record.ts` now measures UTF-8 byte length with `TextEncoder` instead of `node:buffer`. The type-only import of `LogRecord` compiles the DTO source under the client's browser tsconfig, so the DTO must be browser-safe. Behavior is identical, the DTO's own tests are unchanged, and all dependent packages type-check clean.

Reviewed with /plan-ceo-review, /plan-design-review, and /plan-eng-review (plus two outside-voice passes) before implementation.

Part of the Runner log capture & streaming project.

---
## Summary by cubic
Adds `@shipfox/client-logs`, a UI package for rendering process and system step logs with `@shipfox/react-ui`. Makes `@shipfox/api-logs-dto` browser‑safe via `TextEncoder`, and adds `formatBytes`/`formatDuration` to `@shipfox/react-ui`.

- **New Features**
  - `LogView` renders logs with empty state, timestamp modes (`off`/`rel`/`abs`), optional wrap, line numbers, and a stable per‑node `seq`.
  - Record components: `OutputLogRow` (stderr gets a subtle left rule), `LogGroup` (collapsible with duration/running/incomplete and an inset error bar), and markers for `end`/`gap`/`capped`/`runner_lost` with clear copy and right‑aligned meta; excludes `agent_session`.
  - `buildLogTree(records)` reconstructs groups via `group_id`/`parent_group_id`, closes orphans when ends drop, places markers, numbers lines, computes `originTs`/`terminated`/per‑group line counts, and sets `hasError` only on `runner_lost`. Includes unit tests and Storybook + Argos.

- **Refactors**
  - `@shipfox/api-logs-dto`: use `TextEncoder` for UTF‑8 byte length so the DTO compiles in the browser; behavior unchanged.
  - `@shipfox/react-ui`: add `utils/format-bytes` and `utils/format-duration` with tests; re‑export from `utils/index`.

<sup>Written for commit 878c6160db84f4fb3c460f75eadb33fe43fbdc90. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/298?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/client-logs-record-components.md
+++ b/.changeset/client-logs-record-components.md
@@ -1,0 +1,12 @@
+---
+"@shipfox/client-logs": minor
+"@shipfox/react-ui": minor
+"@shipfox/api-logs-dto": patch
+---
+
+Add `@shipfox/client-logs`: the record components for the step-log read stream, composing the `@shipfox/react-ui` log primitives. This covers every process and system record (`output`, `group_start`/`group_end`, `end`, `gap`, `capped`, `runner_lost`); `agent_session` is rendered by the agent-sessions surface.
+
+- `buildLogTree(records)` is a pure transform that reconstructs the group tree from the flat record list. `group_end` closes the matching `group_id` (so a `group_start` dropped under gap/backlog pressure does not mis-nest), record dispatch is an exhaustive switch, and each group node carries a precomputed `hasError` (a `runner_lost` in its subtree, a genuine failure; `stderr` is a channel, not an error) and subtree line count.
+- `OutputLogRow` renders stdout/stderr (stderr gets a subtle left channel rule, not a background tint), `LogGroup` is a collapsible disclosure with running/duration/incomplete affordances and an inset error bar, the system markers render as timeline rows, and `LogView` is the top-level dispatcher with an empty state. Reviewed in a package-local Storybook captured by Argos (`client-logs`).
+- `@shipfox/api-logs-dto` now measures UTF-8 byte length with `TextEncoder` instead of `node:buffer`, so this shared record contract is browser-safe for the client log viewer. Behavior is identical.
+- `@shipfox/react-ui` gains two shared formatters in `utils`: `formatBytes` (new) and `formatDuration` (an ms-span, sub-second sibling to the existing `humanDuration`), so `client-logs` and future packages share one implementation instead of re-rolling them.
--- a/libs/api/logs-dto/src/schemas/record.ts
+++ b/libs/api/logs-dto/src/schemas/record.ts
@@ -1,4 +1,3 @@
-import {Buffer} from 'node:buffer';
 import {z} from 'zod';
 
 /**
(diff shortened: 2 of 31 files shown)


noe-charmet:
Addressed the cubic review (commit 878c6160):

- **P2 — `originTs` anchored to a skipped `agent_session`** (`log-tree.ts`): fixed. `buildLogTree` now anchors `originTs` on the first non-`agent_session` record, so relative timestamps baseline on the first render-relevant row. Test added.
- **P3 — `formatDuration` could render `1000ms`** (`react-ui/utils/duration.ts`): fixed. A fractional value just under 1000 now carries to `1s`. Test added.
- **P3 — `stripTrailingNewline` left a trailing `\r` on CRLF** (`log-tree.ts`): fixed. Now strips `\r?\n$` (top-level regex, per the `useTopLevelRegex` lint). CRLF test added.
- **P3 — `tsconfig.test.json` excludes stories** (vs `exclude: []`): not applied. I tried `exclude: []`, but type-checking the stories then fails: the components have required props and these stories use a custom `render` without `args`, which strict CSF typing rejects. This is the same reason the only other client package with component stories — `@shipfox/client-workflows` — also excludes `**/*.stories.tsx`; the `exclude: []` packages have no such stories. So `client-logs` intentionally matches `client-workflows` here.

Green across `turbo type type:emit check test` for `@shipfox/client-logs` and `@shipfox/react-ui`.

PR #406, opened 2026-06-29
Add per-step source view to the workflow run page

ThierryAbalea:
Adds an icon-only "View source" action on the far right of each located step
row in the Workflow Run Page. It is reachable while the row is collapsed and
opens the existing workflow source panel highlighting the step's lines (from
the step's `source_location`) and scrolling them into view — so a long
workflow no longer leaves the highlighted lines off-screen at the bottom.

## Why

The run page could already open the whole workflow source (ENG-584) and the
panel could highlight a line range, but there was no per-step affordance and
no scroll, so on a long file the highlighted lines were invisible. This closes
that gap and makes "show me where this step is defined" a one-click,
row-level action.

## Design notes

- **Source focus is page-level state decoupled from row selection**, so opening
  or closing the panel never collapses the expanded step or changes the selected
  job/attempt. The summary and per-step controls keep mutually exclusive
  `aria-expanded` (at most one reports the shared panel expanded), and focus
  returns to whichever button opened the panel.
- **The action is a sibling of the Accordion disclosure**, never nested in its
  trigger/`Header` (button-in-button + interactive-control-in-`<h3>` would be
  invalid). It is a hover/focus-revealed row action per DESIGN.md §10, but stays
  keyboard-reachable and is forced visible while its panel is open.
- **One control per step**: a pure step-list-model rule marks the current-attempt
  "anchor" row, so a multi-attempt step renders the Source action once.
- **Scroll-into-view is the one shared change**: an opt-in `scrollHighlightedIntoView`
  prop on the react-ui `CodeBlockContent` primitive, because only it observes when
  the async Shiki markup (the `.highlighted-line` target) has rendered. It honors
  `prefers-reduced-motion`, is guarded for layout-less environments, and defaults
  off, so existing call sites are unaffected.

## Review focus

- The `aria-expanded` mutual-exclusivity + focus-return wiring in `WorkflowRunView`.
- The single react-ui change (the opt-in scroll prop) and its DESIGN.md §14 entry.
- **Highlight precedence (please confirm):** the explicit per-step focus wins, and
  the prior selection-derived highlight is kept as an additive fallback so deep
  links still pre-highlight when the summary opens the panel. Happy to drop the
  fallback if you'd rather the per-step button be the only highlight source.

No backend/DTO changes — `source_location` and `source_snapshot` already exist.

Refs ENG-650

Builds on ENG-460 (backend `source_location` DTO) and ENG-584 (the source panel).

---
## Summary by cubic
Adds per-step and job-level “View source” to the Workflow Run page, focusing the source panel on a step’s lines and scrolling them into view. Refs ENG-650; builds on ENG-460 (`source_location`) and ENG-584.

- **New Features**
  - Workflow Run: “View source” in expanded step details and in the job header for the selected attempt; shown only when a step has a `source_location`; all buttons control the same panel with mutually exclusive aria-expanded.
  - State and a11y: source focus is page-level and independent of row/job selection; focus returns to the opener (job, step, or summary) on close and falls back to the summary button if the opener unmounts; if a focused step loses its location after refetch, the panel downgrades to the whole workflow; per-step focus overrides URL-selected highlight while deep links still pre-highlight via the summary.
  - `@shipfox/react-ui`: `CodeBlockContent` adds `scrollHighlightedIntoView`; the workflow source panel opts in to center the first highlighted line and honors `prefers-reduced-motion`; default remains off elsewhere.

- **Refactors**
  - StepList expanded slot now receives `stepLabel` and `sourceLocation`; job header synchronizes source state with the selected attempt; tests and stories updated, including a JobCard story showcasing the header “View source” action.

<sup>Written for commit d45cfcd04f8be83cb7a391902893e22d5e3d7969. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/406?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DESIGN.md
```diff
@@ -475,3 +475,4 @@ Tags / pills have their own token family because they need distinct background +
...
+| 2026-06-28 | Per-step "View source" is an icon-only, hover/focus-revealed row action on the step row's far right (Remix `codeSSlashLine`, `transparentMuted`); the in-panel highlight scrolls the step's lines to center | Follows §10's "inline row actions revealed on hover" so the affordance stays quiet until needed, while remaining keyboard-reachable (revealed on `focus-visible`, never `display:none`) and forced visible while its panel is open. The button is a sibling of the Accordion disclosure (not nested in its trigger/heading) to keep valid HTML. One control per step (the current-attempt "anchor" row, computed in the pure step-list model) keeps the shared panel's `aria-expanded` mutually exclusive across the summary and step buttons. Scroll-to-highlight is an opt-in `scrollHighlightedIntoView` on the react-ui `CodeBlockContent` primitive because only it observes when the async Shiki markup (the `.highlighted-line` target) has rendered; it reads `prefers-reduced-motion` at scroll time and is guarded so layout-less environments never throw. |
```
noe-charmet: I don't think that this is an absolute rule that should be in our design context forever

libs/client/workflows/src/components/workflow-step-list/workflow-step-list.tsx
```diff
@@ -280,18 +336,27 @@ function WorkflowStepRow({
...
-      )}
+      {/* `group/row` drives the Source action's hover/focus reveal without
+          clashing with the trigger's own `group`. The button is a sibling of the
+          trigger (never inside its Header/button), so it stays valid + focusable. */}
+      <div className="relative group/row">
+        {triggerNode}
+        {sourceButton ? (
+          <div className="absolute right-8 top-1/2 -translate-y-1/2">{sourceButton}</div>
+        ) : null}
+      </div>
```
noe-charmet: I don't think placing the button there is the best idea in terms of UI design.
- The row is already a collapsible trigger. Adding a button on the same surface makes affordance and behavior weird.
- Accessing the source lines of a job's steps is not a top priority action that it should always be displayed even when the row details are hidden.

IMO it's a lot better to start adding an actions row inside the details surface rather than directly on the collapsible trigger. Will make affordance of the button way more clear, and will properly place it's importance level.

libs/client/workflows/src/components/workflow-step-list/workflow-step-list.tsx
```diff
@@ -48,6 +58,17 @@ export interface WorkflowStepListProps {
...
+  // present, the anchor row of each located step renders a Source button.
+  sourcePanelId?: string | undefined;
+  sourceAvailable?: boolean | undefined;
+  focusedSourceStepId?: string | null | undefined;
+  onOpenStepSource?:
+    | ((
+        request: WorkflowStepSourceRequest,
+        triggerRef: RefObject<HTMLButtonElement | null>,
+      ) => void)
+    | undefined;
```
noe-charmet: Looks like a lot of state going up and down the tree. Could we have reached a level where we should add a `jotai` store for this part of the state ?

noe-charmet:
Lot of "LLM comments", I suggest you as Codex something like "Go through all comments added and modified within this PR and ensure they adhere to our guidelines", ChatGPT is pretty good at it.

DESIGN.md
ThierryAbalea: Removed that DESIGN.md note in 33ad3409; agreed it was too absolute for permanent design guidance.

libs/client/workflows/src/components/workflow-step-list/workflow-step-list.tsx
ThierryAbalea: Moved the per-step source action into the expanded details area in c19bd37d, so the collapsed row remains a single accordion trigger.

libs/client/workflows/src/components/workflow-step-list/workflow-step-list.tsx
ThierryAbalea: Kept this local for now: the state only coordinates this run detail/source panel interaction, and moving it to Jotai would widen the surface without current cross-view reuse. We can extract it if this grows beyond the view.

ThierryAbalea:
While touching this area, I also fixed the two cubic findings: per-step source controls now have step-specific accessible names, and closing the source panel now falls back to the summary View source button when a refetch removes the original step trigger.

noe-charmet (dismissed):

PR #410, opened 2026-06-29
[client-workflows] Show job durations on the workflow run job graph

ThierryAbalea:
Each job node on the Workflow Run Page graph now shows a duration that stays honest across phases: a live queue timer while the job waits for a runner, a live execution timer while it runs, and the final span once it finishes.

Both clocks are derived **client-side** from the job's existing `queued_at` / `started_at` / `finished_at` timestamps (wait = `now − queued_at`, execution = `(finished_at ?? now) − started_at`), so there is **no backend or DTO change**. The branchy "which clock, anchored when" decision is a pure `jobDurationDisplay` mapper; the node just renders and ticks.

The value is a single bare, muted, monospace number with **no visible verb** — the status glyph already names the state, keeping one status affordance per dense node. The verb lives in the screen-reader `aria-label` ("queued 6m", "ran 2m 14s"). When the relevant anchor is absent — `skipped`, cancelled-before-dispatch, terminal without `finished_at`, or timestamps not yet projected (the runner queue/claim events are eventually consistent) — the node shows **nothing** rather than a placeholder.

A shared ticker advances live values every second (paused while the tab is hidden), slowing to a calm 10s cadence under `prefers-reduced-motion` so the timer keeps advancing without per-second motion, rather than freezing between polls.

### Review notes
- The duration semantics were designed against how GitHub Actions presents the same thing, and the implementation was vetted by an independent consult review; its findings (test coverage of the live tick + reduced-motion, doc accuracy) are applied.
- Coverage: pure mapper node tests, ticker/label dom tests, and a `WithDurations` Storybook story captured in light + dark via the existing `client-workflows` Argos build (anchors sit near Storybook's frozen `Date.now()` so snapshots stay deterministic).
- Deferred (with triggers, noted in the issue): the `aria-label` spoken value can lag the visible digits between re-renders (only matters if the node becomes a live region); reconciling DESIGN.md §9's 11-state taxonomy with the real 6-value job enum.

Closes ENG-651

---
## Summary by cubic
Show per-job durations on the workflow run job graph with live queue/run timers and final spans. Durations are projected by the workflow API and update live via a shared ticker (closes ENG-651).

- **New Features**
  - Duration source: API attaches a `duration` descriptor from timestamps (`queued`/`running`/`finished`/`none`); `none` for skipped, cancelled-before-start, or missing anchors; skewed finished spans clamp to 0s. Exposes `JobDurationDto`/`jobDurationDtoSchema` in `@shipfox/api-workflows`/`@shipfox/api-workflows-dto` and maps via `jobDurationFor`.
  - Live updates: Shared `TimeTickerProvider`/`useTimeTick` in `@shipfox/react-ui` ticks every second, pauses when hidden, and shifts to 10s under reduced motion. `RelativeTime` now uses this ticker; job graph labels use `humanDuration`, and visible digits and the button `aria-label` stay in sync.

- **Bug Fixes**
  - Stabilized Source Control Onboarding test by awaiting the “Install GitHub” link.
  - Fixed workflow client DTO type references for job duration fields.

<sup>Written for commit 5bafa8db31443e2a476eed908b50061aafb6612f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/410?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libs/client/workflows/src/components/workflow-jobs-graph/job-duration.ts
```diff
@@ -0,0 +1,37 @@
...
+    // Terminal without a finish timestamp (crash/lag): a live clock would be wrong
+    // and a span is unknown, so show nothing.
+    return {kind: 'none'};
+  }
+
+  // Never started executing.
+  if (terminal) return {kind: 'none'};
+  if (queuedAt !== null) return {kind: 'queued', fromIso: queuedAt};
+  return {kind: 'none'};
+}
```
noe-charmet: I feel like this is not exclusive to the job graph. Could we pull this data back up inside the job core model ?

libs/client/workflows/src/components/workflow-jobs-graph/job-duration-label.tsx
```diff
@@ -0,0 +1,35 @@
+import {Code, humanDuration} from '@shipfox/react-ui';
+import type {JobDurationDisplay} from './job-duration.js';
+import {useDurationTick} from './job-duration-ticker.js';
+
+/**
+ * Renders a job node's duration as a bare, muted, monospace number. There is no
+ * verb: the node's status icon already names the state, so `6m` beside a queued
+ * job and `2m 14s` beside a succeeded one read unambiguously. The verb-bearing
+ * phrasing lives in the node's `aria-label` for screen readers.
+ */
```
noe-charmet: 🤖 🗨️

libs/client/workflows/src/components/workflow-jobs-graph/job-duration-ticker.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const query = window.matchMedia('(prefers-reduced-motion: reduce)');
+    const onChange = () => setReduced(query.matches);
+    query.addEventListener('change', onChange);
+    return () => query.removeEventListener('change', onChange);
+  }, []);
+
+  return reduced;
+}
+
+export function JobDurationTickerProvider({children}: {children: ReactNode}) {
```
noe-charmet: We have libs/shared/react/ui/src/components/relative-time/relative-time.tsx which is kind of supposed to do the same thing.

I think we can align them as a single component in the UI library with the behavior of the one implemented here. Would avoid re-implementing the same piece of utility at multiple places.

libs/client/workflows/src/components/workflow-jobs-graph/job-duration.ts
ThierryAbalea: Done. The duration projection now lives in workflow core/API (`jobDurationFor`), and the client consumes the DTO field instead of deriving it in the graph.

libs/client/workflows/src/components/workflow-jobs-graph/job-duration-ticker.tsx
ThierryAbalea: Done. The live duration ticker is shared through `@shipfox/react-ui`, so the graph no longer owns a duplicate ticker implementation.

libs/client/workflows/src/components/workflow-jobs-graph/job-duration-label.tsx
ThierryAbalea: Handled in the follow-ups. The label rendering is now explicit/exhaustive, and the durable UI rule is captured in `DESIGN.md` instead of narrating the implementation in source comments.

ThierryAbalea:
@noe-charmet I pushed the follow-ups: duration projection is in workflow core/API, the live ticker is shared through react-ui, label rendering is exhaustive, and the `DESIGN.md` entry is trimmed to the durable UI rule. Could you take another look?

PR #422, opened 2026-06-29
[api/definitions] Fix VCS sync failing when a manual definition shares a config_path

ThierryAbalea:
Once a manual (or validated) workflow definition existed at a `config_path`, VCS sync of that same path failed with a Postgres unique violation (`23505`), so the workflow never synced and pushes produced no runs. Found via an onboarding first-run simulation exercising the real "validate then push" path.

## Root cause

`definitions_wd_project_id_config_path_unique` was **source-agnostic** — one row per `(project_id, config_path)` across manual *and* VCS definitions. But VCS definitions are versioned by `ref` (see `applyVcsDefinitionsBatch`), so the VCS upsert's `ON CONFLICT` arbiter is the ref index, not the config_path index. When a manual row already held the config_path slot, the VCS insert violated a unique index that was **not its conflict arbiter** → unhandled `23505`. The same flaw would also collide two branches syncing the same `config_path`.

## Fix

Scope the index (and the manual upsert's `ON CONFLICT` predicate) to manual rows only (`config_path IS NOT NULL AND ref IS NULL AND sha IS NULL`), so manual and VCS definitions coexist per `config_path`. Adds migration `0002` and a regression test.

## Review notes

- Migration `0002` drops + recreates the partial unique index. `drizzle-kit generate` couldn't run in my worktree (workspace export-condition resolution), so it's hand-written in the repo's existing style (0001 is also hand-written; no per-migration snapshots are kept). The test setup re-applies migrations via `runMigrations`, so the suite exercises it.
- Worth a second look: confirm no read path assumes one definition per `config_path` regardless of source. I checked the obvious ones — resolution is by definition id, `applyVcsDefinitionsBatch` reads by `(project, ref, config_path)`, `listDefinitionsByProject` returns all — and `@shipfox/api-definitions` tests are green.

Verified locally: `@shipfox/api-definitions` test **202/202**, `turbo type`/`check` green, `turbo build`/`check` green across affected packages. The full `turbo test --filter '...[origin/main]'` surfaced 4 failures unrelated to this change (`react-ui`, `client-workflows`, `api-logs` — none depend on api-definitions — and `api-workflows` temporal integration tests); the worktree was 19 commits behind `origin/main`, inflating the affected set. CI runs the authoritative gate.

Fixes ENG-659

---
## Summary by cubic
Fix VCS sync failing when a manual definition shares a config_path by scoping the unique index to manual rows and enforcing source/ref/sha consistency. Manual and VCS definitions now coexist per path so pushes create runs as expected. Addresses ENG-659.

- **Bug Fixes**
  - Scoped the config_path unique index to manual rows (`ref`/`sha` null) and matched the manual upsert `ON CONFLICT` predicate; VCS rows use `ref`/`sha` indexes so branches and manual entries no longer collide.
  - Bound `source` to git coordinates via a DB CHECK constraint and validation in `@shipfox/api-definitions` and `@shipfox/api-definitions-dto` (VCS requires `ref` or `sha`; manual requires neither). Added tests for manual+VCS coexistence, two branches, and invalid cases.

- **Migration**
  - Drops/recreates the partial unique index (now manual-only) and adds the CHECK constraint. Run migrations; no app changes needed.

<sup>Written for commit 82d44894d121cff8d2459690b65dbcbf68d03f47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ShipfoxHQ/shipfox/pull/422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libs/api/definitions/src/db/definitions.test.ts
+++ b/libs/api/definitions/src/db/definitions.test.ts
@@ -130,6 +130,7 @@ describe('definition queries', () => {
         configPath: 'ci.yml',
         name: 'CI',
         ...definitionFields('CI'),
+        source: 'vcs',
         sha: 'abc123',
       });
 
@@ -144,6 +145,7 @@ describe('definition queries', () => {
         configPath: 'ci.yml',
         name: 'CI v1',
         ...definitionFields('CI v1'),
+        source: 'vcs',
         sha: 'abc123',
       });
 
@@ -153,6 +155,7 @@ describe('definition queries', () => {
         configPath: 'ci.yml',
         name: 'CI v2',
         ...definitionFields('CI v2'),
+        source: 'vcs',
         sha: 'abc123',
       });
 
@@ -167,6 +170,7 @@ describe('definition queries', () => {
         configPath: 'ci.yml',
         name: 'CI',
         ...definitionFields('CI'),
+        source: 'vcs',
         ref: 'main',
       });
 
@@ -181,6 +185,7 @@ describe('definition queries', () => {
         configPath: 'ci.yml',
         name: 'CI v1',
         ...definitionFields('CI v1'),
+        source: 'vcs',
         ref: 'main',
       });
 
@@ -190,6 +195,7 @@ describe('definition queries', () => {
         configPath: 'ci.yml',
         name: 'CI v2',
         ...definitionFields('CI v2'),
+        source: 'vcs',
         ref: 'main',
       });
 
@@ -220,6 +226,57 @@ describe('definition queries', () => {
       expect(vcs.id).not.toBe(manual.id);
       expect(manual.source).toBe('manual');
       expect(vcs.source).toBe('vcs');
+    });
+
+    test('VCS definitions on two branches coexist at the same config_path', async () => {
+      const main = await upsertDefinition({
+        projectId,
+        workspaceId,
+        configPath: 'ci.yml',
+        name: 'CI main',
+        ...definitionFields('CI main'),
+        source: 'vcs',
+        ref: 'main',
+      });
+
+      const dev = await upsertDefinition({
+        projectId,
+        workspaceId,
+        configPath: 'ci.yml',
+        name: 'CI dev',
+        ...definitionFields('CI dev'),
+        source: 'vcs',
+        ref: 'dev',
+      });
+
+      expect(dev.id).not.toBe(main.id);
+    });
+
+    test('rejects a vcs definition with neither ref nor sha', async () => {
+      const upsert = upsertDefinition({
+        projectId,
+        workspaceId,
+        configPath: 'ci.yml',
+        name: 'CI',
+        ...definitionFields('CI'),
+        source: 'vcs',
+      });
+
+      await expect(upsert).rejects.toThrow();
+    });
+
+    test('rejects a manual definition that sets a sha', async () => {
+      const upsert = upsertDefinition({
+        projectId,
+        workspaceId,
+        configPath: 'ci.yml',
+        name: 'CI',
+        ...definitionFields('CI'),
+        source: 'manual',
+        sha: 'abc123',
+      });
+
+      await expect(upsert).rejects.toThrow();
     });
 
     test('with neither sha nor ref uses base constraint', async () => {
@@ -516,6 +573,7 @@ describe('definition queries', () => {
         configPath: 'ci.yml',
         name: 'CI',
         ...definitionFields('CI'),
+        source: 'vcs',
         ref: 'main',
       });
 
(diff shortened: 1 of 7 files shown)


libs/api/definitions/drizzle/0002_scope_config_path_unique_to_manual.sql
noe-charmet, line 1: You can edit the existing migration as we're pre-deploy

libs/api/definitions/drizzle/0002_scope_config_path_unique_to_manual.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX IF EXISTS "definitions_wd_project_id_config_path_unique";--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS "definitions_wd_project_id_config_path_unique" ON "definitions_workflow_definitions" USING btree ("project_id","config_path") WHERE "config_path" IS NOT NULL AND "ref" IS NULL AND "sha" IS NULL;
```
noe-charmet: Why the `"ref" IS NULL AND "sha" IS NULL`.
This would only impact the API uploaded workflows and not the synced ones it seems

libs/api/definitions/src/db/definitions.test.ts
```diff
@@ -197,6 +197,37 @@ describe('definition queries', () => {
       expect(second.name).toBe('CI v2');
     });
 
+    // ENG-659 regression: a manual definition and a VCS-synced definition (keyed by
+    // ref, see applyVcsDefinitionsBatch) at the SAME (projectId, configPath) must
+    // coexist. The config_path unique index is scoped to manual rows (no ref, no sha),
+    // so the VCS upsert no longer collides on an index that isn't its ON CONFLICT
+    // arbiter. Before the fix this rejected with a unique violation on
+    // definitions_wd_project_id_config_path_unique.
```
noe-charmet: Not on our commenting guidelines

libs/api/definitions/drizzle/0002_scope_config_path_unique_to_manual.sql
ThierryAbalea: The crux is how Postgres `ON CONFLICT` works: it only resolves a collision on **the single index it names** (the arbiter). If the insert trips a *different* unique index, it doesn't update — it throws `23505`.

Our upsert picks the arbiter by `sha`/`ref` presence (not `source`): a manual write (no ref/sha) → `config_path_unique`; a sync write (has `ref`) → `ref_lookup`. The bug: `config_path_unique` was `WHERE config_path IS NOT NULL` (source-agnostic), so —

```
manual row:  (P, ci.yml, ref=NULL, sha=NULL)  → sits in config_path_unique
push→sync:   (P, ci.yml, ref=main)            → arbiter is ref_lookup, but the
             INSERT also hits the old config_path_unique (a manual row is there)
             → that's not its arbiter → unhandled 23505 → sync dies
```

So the predicate has to describe **the rows the arbiter targets** — "no ref, no sha" — not `source`. `ref IS NULL AND sha IS NULL` excludes the sync row (it has a `ref`), so it stops tripping it. And yes, it now only constrains manual rows — intentional: VCS uniqueness already lives in `ref_lookup` / `sha_lookup` (per branch / per commit); putting VCS rows in `config_path_unique` is what made sync collide.

(`source = 'manual'` *looks* equivalent but isn't — the arbiter ignores `source`, so a `vcs` row with no ref/sha would slip through.) The follow-up commit adds a CHECK binding `source ⟺ ref/sha` (+ request validation + negative tests), so `manual ⟺ ref IS NULL AND sha IS NULL` is now enforced and the predicate choice is provably correct.

libs/api/definitions/drizzle/0002_scope_config_path_unique_to_manual.sql
ThierryAbalea: Done — folded the index change into `0000_initial.sql` and dropped `0002` (pre-deploy).

libs/api/definitions/src/db/definitions.test.ts
ThierryAbalea: Done — removed the comment; the explanation now lives in the changeset and commit message.

noe-charmet (dismissed):

noe-charmet (approved):

PR #1172, opened 2026-07-30
Generate the workflow contexts reference from the expression registry

noe-charmet:
Splits the expressions reference in two and generates the new page from the engine, so the reference cannot drift from what a workflow can actually read.

`reference/expressions` now covers only syntax, types, operators, and functions, with a working example on every row. `reference/contexts` is new and holds every context, its properties, and the availability matrix keyed by the YAML field an author types. Both tables and every property description are generated from `@shipfox/expression`, and `apps/docs/scripts/check-contexts.mjs` fails the build when the engine gains an expression site the docs do not name, when a property has no description, or when a description outlives its property. That check already caught the checkout target fields added by #1185.

Two product changes come with it, each with its own changeset.

`workflow` becomes a context root holding `workflow.name` and `workflow.id`, replacing `run.workflow_name` and `run.definition_id`. `run` now holds instance facts only, so `workflow` and `run` mirror the `job` and `execution` pair one level down and the generated tables teach one rule instead of a rule plus an exception. This removes fields from an exported type environment, so `@shipfox/expression` takes a major bump.

`agent.thinking` accepts a `${{ }}` interpolation. The published JSON Schema keeps the per-harness enum for editor completion through an `anyOf` branch, and the resolved value is checked against the harness levels when the step dispatches. An unsupported level fails the step rather than falling back, so a workflow never silently runs at a different reasoning effort than it asked for.

The rest is placement. The shell re-evaluation warning moved from the concept page to the `run` field it constrains and now lists all eleven constructs the sync warns about. Recovery from a context that is not available moved to the workflow sync guide, since it is a definition error rather than a runtime one. The glossary entry for CEL stopped restating gate fields.

Review notes. ENG-1454 is resolved upstream by #1178 and #1181, which is what unblocked generating the matrix from the registry rather than hand-authoring it. The security callout is the first `type="warn"` callout in the docs; fumadocs-ui 16.7.10 supports it, and it is one word to change if the repo prefers `info` everywhere.

Closes ENG-1414

### DIFF
--- a/.changeset/agent-thinking-interpolation.md
+++ b/.changeset/agent-thinking-interpolation.md
@@ -1,0 +1,14 @@
+---
+"@shipfox/workflow-document": minor
+"@shipfox/api-definitions": minor
+"@shipfox/api-definitions-dto": minor
+"@shipfox/api-workflows": minor
+"@shipfox/api-workflows-dto": minor
+"@shipfox/api-agent": minor
+"@shipfox/api-agent-dto": minor
+---
+
+Accept a `${{ }}` interpolation in an agent step's `thinking` field. The schema
+still offers the per-harness enum for editor completion, and the resolved value
+is checked against the harness levels when the step dispatches. An unsupported
+resolved level fails the step.
--- a/.changeset/workflow-context-root.md
+++ b/.changeset/workflow-context-root.md
@@ -1,0 +1,11 @@
+---
+"@shipfox/expression": major
+"@shipfox/api-workflows": minor
+---
+
+Split workflow definition facts out of the `run` context into a `workflow` root.
+`run.workflow_name` becomes `workflow.name` and `run.definition_id` becomes
+`workflow.id`, so `workflow` and `run` mirror the `job` and `execution` pair.
+Adds `contextRootsForField`, which returns the readable roots for a predicate or
+an interpolation field without the caller choosing a mechanism, and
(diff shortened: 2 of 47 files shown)


noe-charmet:
Addressed both cubic findings in cdfff0652.

**Shell code positions (workflow-schema.mdx): valid, fixed.** The callout named six constructs while `shell-code-position.ts` flags eleven, so it did imply `source "$VALUE"` and `bash -c "$VALUE"` were safe. It now lists the full enforced set (`eval`, `sh -c`, `bash -c`, `source`, `let`, `declare -i`, shell arithmetic, `awk`, `jq`, `sed`, `xargs sh -c`) and states that a workflow sync warns when it finds one, which the previous text never mentioned.

**Cross-type comparison (expressions.mdx): not valid for this product.** The comment is right about the CEL spec, where `==` across differing types yields `false`. Our engine is `@marcbachmann/cel-js`, and it raises an overload error instead:

```
1 == "1"  => no such overload: int == string
1 != "1"  => no such overload: int == string
"a" < 1   => no such overload: string < int
1 == 1.0  => no such overload: int == double
```

So the sentence is accurate as written for Shipfox, and softening it to match the spec would tell readers that a mismatched equality quietly returns `false` when it actually fails the expression. I kept the claim and added the `int` versus `double` case, which is the one most likely to surprise an author comparing a numeric output against a literal.

PR #1463, opened 2026-08-25
[client-workflows] Remove duplicate job log divider

noe-charmet:
Removes the redundant divider below the job detail header so it no longer sits directly against the logs panel border. The panel remains the single boundary, keeping the job view compact while reducing visual weight.

Validation:
- `mise exec -- turbo test --filter=@shipfox/client-workflows` (635 tests)
- `mise exec -- turbo check type build depcruise --filter=@shipfox/client-workflows...`
- Impeccable layout scan

### DIFF
--- a/libs/client/workflows/src/components/job-detail/job-detail-header.test.tsx
+++ b/libs/client/workflows/src/components/job-detail/job-detail-header.test.tsx
@@ -3,21 +3,6 @@ import {workflowJob, workflowJobExecutionDto} from '#test/fixtures/workflow-run.
 import {JobDetailHeader} from './job-detail-header.js';
 
 describe('JobDetailHeader', () => {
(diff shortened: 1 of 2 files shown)


libs/client/workflows/src/components/job-detail/job-detail-header.test.tsx
```diff
@@ -3,6 +3,21 @@ import {workflowJob, workflowJobExecutionDto} from '#test/fixtures/workflow-run.
 import {JobDetailHeader} from './job-detail-header.js';
 
 describe('JobDetailHeader', () => {
+  test('does not draw a divider below the header', () => {
```
noe-charmet: Do not add a test that assess absence of a design. We do not add tests that assess that a design is not something.

